Issue #38, opened 2026-02-05
Minute strings are not padded with zeros


tjasz:
The minute strings are not padded with zeros, making the times in a schedule difficult to read. Ex: "09:08" is written as "9:8". This is true for the current time of day as well as the scheduled arrival times.

<img width="485" height="372" alt="Image" src="https://github.com/user-attachments/assets/760512b3-deec-4dc7-a345-c0fed6c9b6f9" />

For consistency, a helper function should be added to format time given an hour and minute. It should include `.padStart(2,'0')` on the minutes. It should possibly also include it for the hour as well, depending on the project's preference.

Then, the helper function should be used in the following two places
- [Line 212 in the Schedule Preview component](https://github.com/organicmaps/gtfs-osm-matcher/blob/main/src/uielements/schedule-preview.tsx#L212) for the current time
- [Line 189](https://github.com/organicmaps/gtfs-osm-matcher/blob/main/src/uielements/schedule-preview.tsx#L189) for the scheduled times

PR #53, opened 2026-02-19
Format hours in padded 24h format in the range 00-23

### LINKED ISSUE #45: Fix displayed time in schedules
Two issues:
1. Print 00, 01.. instead of 24, 25: `24:11 24:26 24:28 24:38 24:43`
2. Print 23:09 instead of `23:9`

tjasz:
Prints "00:08" instead of "24:08" for trips that cross midnight. Prints "01:30" instead of "25:30" similarly.

Prints "07:08" instead of "7:08".

The zero-padding is necessary for consistency. Most people would not format 8 minutes after midnight as "0:08". They would use "00:08" or "12:08a". The 24-hour format with padded zeroes is less complex, less verbose, and familiar to developers and people from most of the world.

Before:
before

After:
after

Builds on #38 #39

Fixes #45


biodranik (approved):
Thanks! As this web tool is for osmers/developers, it's okay to normalize the time )

PR #58, opened 2026-02-20
Fetch node members of a way from OSM if not present in data

### LINKED ISSUE #55: Undefined lonLat from OSMData causes exception
Repro:
1. Go to https://gtfs-osm-matcher.organicmaps.app/#/match-report/seattle-sound-transit/selection/no-match/990008
2. Click the other unmatched stop (on the SW side of Rainier Ave S) and observe that the left overlay info does change.
3. Observe the exception from the screenshot below in the console.
4. Click on the first unmatched stop (on the NE side of Rainier Ave S) again. The left overlay info does not change. You get into a state where clicking any stop does not result in updates to the overlay info until you refresh the page.

Image

Despite the code obfuscation, you can tell that it is coming from [these lines](https://github.com/organicmaps/gtfs-osm-matcher/blob/737c6de157d147fdad8a26343a7f437ab1c24789/src/uielements/selection-info.tsx#L268C1-L279C8) and occurs when `lonLat` is `undefined`.

I have not yet figured out *why* you get `undefined` here (i.e. why you fail to load these nodes).

Also, TypeScript should be catching this undefined dereference at build time. But the warning is suppressed with the `!` operator.

tjasz:
1. `OsmQuerryQueue.ts`: After fetching ways, extract all referenced node IDs and fetch them via a second `/nodes.json` call (skipping nodes already loaded). This allows `getElementLonLat()` to properly compute way centroids.
2. `selection-info.tsx` - `osmMapElements`: Use `OSM_DATA.getLonLat(f)` instead of `f.lon`/`f.lat` directly. Skip rendering the marker if `lonLat `can't be resolved (return `null`).
3. `selection-info.tsx` - `OsmListElement`: Resolve the feature's lon/lat via `OSM_DATA.getLonLat(osmFeature)` with a fallback to `f.lon`/`f.lat` from report data.

This resolves #55 

Before/after of an area around Mt Baker light rail station and Mt Baker (bus) transit center, as described in the bug.

Before:
image

After:
image

The platforms and station centroids now appear at Mt Baker Station (SW-ish area of the screen shot).

There are no longer exceptions thrown for undefined coordinates that crash the rendering of the selection details overlay.

### DIFF
--- a/src/services/OsmQuerryQueue.ts
+++ b/src/services/OsmQuerryQueue.ts
@@ -102,6 +102,28 @@ class OSMQueryQueue {
                     import.meta.env.DEV
                         && console.log('osm ways data', data);
                     OSM_DATA.updateOverpassData(data);
+                    return data;
+                })
+                .then(data => {
+                    // Ways reference node IDs but don't include node coordinates.
+                    // Fetch any referenced nodes that are not already in OSM_DATA
+                    // so we can compute way centroids.
+                    const wayNodeIds = (data.elements || [])
+                        .filter((e: any) => e.type === 'way' && Array.isArray(e.nodes))
+                        .flatMap((e: any) => e.nodes)
+                        .filter((nid: number) => !OSM_DATA.getNodeById(nid));
+
+                    const uniqueNodeIds = [...new Set<number>(wayNodeIds)];
+
+                    if (uniqueNodeIds.length > 0) {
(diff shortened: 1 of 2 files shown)


tjasz:
@kiselev-dv Thanks. I may take a look.

kiselev-dv:
Thank you

PR #67, opened 2026-02-20
Add gh workflow to build the project

kiselev-dv:


.github/workflows/build.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Enable Corepack
+        run: corepack enable
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
biodranik: Why 20 and not 24?

PR #133, opened 2026-06-06
Fixed preview decoding, cancel update requests after 3 errors

kiselev-dv:


src/services/ScheduleEncoding.ts
```diff
@@ -324,7 +324,11 @@ export function servicePeriodIndexes(periods: Periods, date: number): number[] {
   const result: number[] = [];
   const rawBits = decompressRaw(periods);
 
-  if (periods.encoding === "bits" && periods.begin != null && periods.end != null) {
+  // transpose_rle reconstructs per-period blobs WITH a 4-byte begin/end header
+  // (see transposeRleToRawBitsArray), so it must go through decodePeriodBits like
+  // the "begin:end:bits" encoding — not the headerless global-range branch below.
+  if (periods.encoding === "bits" && periods.compression !== "transpose_rle"
```
biodranik, line 330: ✅ **Confirmed correct.** `schedule.types.d.ts:67-70` documents that `transpose_rle` carries `encoding: 'bits'` with hoisted `begin`/`end`, so without this guard the data fell into the headerless global-range branch and the 4-byte per-period header got fed into the bit vector. The encode side (`transposeRleToRawBitsArray`) and `decodePeriodBits` agree on big-endian begin/end + LSB-first bits, so routing transpose_rle to the `else` branch is right.

_Optional follow-up (out of scope here):_ if `decompressRaw` always emitted header-bearing blobs (synthesizing a header from the global begin/end for the headerless `bits` case), `servicePeriodIndexes` would collapse to a single `decodePeriodBits` path and this special-case would disappear. Cleaner, but it changes the documented `decompressPeriods` contract.

src/services/ScheduleEncoding.ts
```diff
@@ -440,14 +444,24 @@ export function dateAsNumber(date: Date) {
   return (date.getFullYear() % 100) * 10000 + (date.getMonth() + 1) * 100 + date.getDate();
 }
 
+/**
+ * Converts a day-number (days since 2000-01-01) to a yyMMdd number.
+ * Uses UTC getters because EPOCH_MS is UTC midnight — using local getters would
+ * shift the calendar date by one day in timezones behind UTC.
+ */
+function dayNumberAsYyMMdd(dayNumber: number): number {
```
biodranik, line 452: ✅ **Confirmed correct.** `EPOCH_MS` is UTC midnight, so the old local getters shifted the calendar date by −1 day in zones behind UTC. Worth noting the intended asymmetry: this calendar conversion is UTC, while the *query* date `dateAsNumber(new Date())` (`i_today` in schedule-preview) correctly stays **local** — the user's today. Both yield `yyMMdd`, so matching a local today against absolute calendar service dates is exactly right.

src/uielements/schedule-preview.tsx
```diff
@@ -39,6 +39,7 @@ export function SchedulePreview({ selection }: SchedulePreviewProps) {
 
     useEffect(() => {
         setSchedules([]);
+        setTripUpdates(undefined);
```
biodranik, line 42: **(C)** `liveUpdates` is not reset when `id` changes, so selecting a non-live stop right after a live one briefly polls `/updates/{newId}` until the new stop's `feedMeta` arrives — a spurious request (often a 404 that also feeds the error counter below). Reset it here, next to `tripUpdates`:
```suggestion
        setTripUpdates(undefined);
        setLiveUpdates(false);
```

src/uielements/schedule-preview.tsx
```diff
@@ -39,6 +39,7 @@ export function SchedulePreview({ selection }: SchedulePreviewProps) {
...
 
         fetch(`${SchedulesAPIBase}/${id}`).then(r => r.json()).then((data: any) => {
```
biodranik: **(B)** This schedule fetch is not guarded by `id`/cancellation. The updates effect below got a `cancelled` flag, but here a slow response for a previously-selected stop can resolve last and overwrite `schedules`/`liveUpdates` for the current selection. Mirror the same guard (or use `AbortController`):
```ts
let cancelled = false;
fetch(`${SchedulesAPIBase}/${id}`)
    .then(r => r.json())
    .then(data => { if (cancelled) return; /* …setSchedules / setLiveUpdates… */ });
return () => { cancelled = true; };
```

src/uielements/schedule-preview.tsx
```diff
@@ -66,16 +67,45 @@ export function SchedulePreview({ selection }: SchedulePreviewProps) {
...
+
+                if (response.status === 200) {
+                    const data = await response.json();
+                    if (cancelled) return;
+                    if (data) {
+                        setTripUpdates(data);
+                    }
+                }
+                else if (errCounter++ > 2) {
+                    clearInterval(rt);
```
biodranik, line 103: **(A)** The breaker counts **cumulative**, not consecutive, failures — `errCounter` is never reset on success, so over a long session a few scattered blips permanently kill live updates. Also `errCounter++ > 2` first trips on the **4th** failure, not the 3rd (the title says 'after 3 errors'). Reset on a 200 and pre-increment:
```suggestion
                    if (data) {
                        setTripUpdates(data);
                    }
                    errCounter = 0;             // reset breaker: a 200 means the endpoint is healthy
                }
                else if (++errCounter >= 3) {  // cancel after 3 consecutive failures
                    clearInterval(rt);
```
The `catch` branch needs the same `++errCounter >= 3` (see next comment).

src/uielements/schedule-preview.tsx
```diff
@@ -66,16 +67,45 @@ export function SchedulePreview({ selection }: SchedulePreviewProps) {
...
+                }
+            }
+            catch {
+                if (!cancelled && errCounter++ > 2) {
```
biodranik: **(A cont.)** Same off-by-one + counter fix here:
```suggestion
                if (!cancelled && ++errCounter >= 3) {
```

biodranik (commented):
## Review — PR #133 (architecture · frontend · TypeScript · perf)

Reviewed both changed files and all callers — `SchedulePreview` is the only consumer, and `decodeScheduleOnDate` its only entry into the encoder. `tsc -b` passes.

**Both fixes are correct** ✅ — details inline on the `transpose_rle` guard and the UTC date helper.

Inline comments cover **(A)** the error-counter threshold/reset, **(B)** the unguarded schedule fetch, and **(C)** the missing `liveUpdates` reset. The items below sit on lines **outside this PR's diff**, so they are here rather than inline:

### (D) `.sort()` mutates memoized data during render — `schedule-preview.tsx:135,139`
`schedule.sort(...)` and `routes.sort(...)` sort the `schedulesPerRegion` `useMemo` result in place, and `regionStops` is unmemoized, so this re-runs on every render — i.e. every 5 s realtime tick re-sorts and rebuilds the whole subtree. No functional bug in Preact today (stable sort, synchronous render), but it is a render-purity violation + wasted work. Sort copies, ideally inside the decode memo:
```ts
const decoded = [...schedule].sort(stopPlatformCodeCmp);
const routeAndTrips = [...routes].sort((a, b) => …);
```

### (E) Duplicate log — `schedule-preview.tsx:53-54`
`console.error('Old timetable format is not supported')` is printed twice; delete one.

### (F) Type misnomer — `schedule-preview.tsx:16`
`ScheduleApiResponseV3` gates on `formatVersion === '4'` (`:49`) — rename to `ScheduleApiResponseV4`.

### Comment accuracy
- `ScheduleEncoding.ts:474` — JSDoc references `decodeRlePeriodBits`, which does not exist; drop/correct it.
- `ScheduleEncoding.ts:492` — `isServiceDay` says 'yyDDmm local date'; it is `yyMMdd`, and after this PR the service-date set is UTC-derived.

### Recommended simplification (optional)
Extract the realtime polling into a `useLiveUpdates(id, enabled)` hook using `AbortController` + recursive `setTimeout`: the recursion removes the `inFlight` guard, `abort()` removes the post-`await` `cancelled` checks, and the failure threshold lives in one place (folding in **A**). Trade-off: cadence becomes '5 s after each request finishes' rather than fixed wall-clock ticks — fine for a delay poll.

PR #144, opened 2026-06-30
Show list of already matched gtfs stops for osm candidate

kiselev-dv:
There is a warning for osm candidates if they are already matched to any GTFS stop. Display list of those stops

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -15,3 +15,6 @@ html, body {
   width: 100%;
 }
 
+.cursor-pointer {
+  cursor: pointer;
+}
--- a/src/uielements/selection-info.tsx
+++ b/src/uielements/selection-info.tsx
@@ -358,8 +358,8 @@ type OsmListElementProps = {
 function OsmListElement({ f, editDefault, parentLonLat, tagActions, mouseEvents }: OsmListElementProps) {
 
     const [edit, setEdit] = useState(editDefault || false);
-
     const [version, setVersion] = useState(0);
+    const [warnExpanded, setWarnExpanded] = useState(false);
 
     const type = f.id[0] === 'n' ? 'node' : 'way';
     const idn = f.id.slice(1);
@@ -380,8 +380,6 @@ function OsmListElement({ f, editDefault, parentLonLat, tagActions, mouseEvents 
 
     const osmUrl = `https://osm.org/${type}/${idn}`;
     const osmHref = <a target="_blank" href={osmUrl}>{f.id}</a>;
-
-    const matchSet = f.matchSet;
 
     const osmFeature = OSM_DATA.getByTypeAndId(type, idn);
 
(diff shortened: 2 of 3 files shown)


src/uielements/selection-info.tsx
```diff
@@ -431,10 +429,18 @@ function OsmListElement({ f, editDefault, parentLonLat, tagActions, mouseEvents
...
-            <span>&#9888;</span>This OSM Element is already matched as {matchSet}
-        </div>;
+    const alreadyMatchWarning = f.mtch && f.mtch.length > 0 && <>
+        <div className={'warning cursor-pointer'} onClick={() => setWarnExpanded(!warnExpanded)}>
+            <i>&#9888; </i>
+            <span>This OSM Element is already matched to another gtfs feature</span>
+        </div>
+        {
+            warnExpanded && <ul>{
+                f.mtch.map((m: string) => {return (<li><a href={`/#/match-report/swiss-opendata/selection/${m}`}>{m}</a></li>)})
```
biodranik: **Bug: hardcoded region + missing `key`**

`swiss-opendata` is hardcoded in the URL, so links are broken for every other dataset. `reportRegion` is already in `MatchInfoProps` (though currently not destructured by `MatchInfo`); it can be threaded as a prop through `OsmElements` → `OsmListElement`, or read directly from `SelectionContext` (already used in sibling components):
```tsx
const { selection } = useContext(SelectionContext);
const reportRegion = selection?.reportRegion;
```

Also, `<li>` is missing a `key` prop (`m` is the natural key), and the map arrow body can drop the redundant `{return ()}` wrapper:
```tsx
f.mtch.map((m: string) => <li key={m}><a href={`/#/match-report/${reportRegion}/selection/${m}`}>{m}</a></li>)
```

---

src/uielements/selection-info.tsx
```diff
@@ -431,10 +429,18 @@ function OsmListElement({ f, editDefault, parentLonLat, tagActions, mouseEvents
         OSM_DATA.setNodeLatLng({ lng: lonLat[0], lat: lonLat[1] }, osmFeature);
     }, [osmFeature]);
 
-    const alreadyMatchWarning = matchSet && matchSet !== 'no-match' &&
-        <div className={'warning'}>
-            <span>&#9888;</span>This OSM Element is already matched as {matchSet}
-        </div>;
+    const alreadyMatchWarning = f.mtch && f.mtch.length > 0 && <>
+        <div className={'warning cursor-pointer'} onClick={() => setWarnExpanded(!warnExpanded)}>
+            <i>&#9888; </i>
```
biodranik, line 437: `<i>` is for italic text / icon fonts; wrapping a bare Unicode character with it is unconventional. A `<span>` (as used before) is more semantically appropriate here, and the cursor style can be applied directly to `.warning` in `selection-info.css` instead of via the global `.cursor-pointer` utility (see `index.css` comment).

---

src/index.css
```diff
@@ -15,3 +15,6 @@ html, body {
   width: 100%;
 }
 
+.cursor-pointer {
+  cursor: pointer;
+}
```
biodranik, line 20: `.cursor-pointer` is only used by `selection-info.tsx`; it should live in `selection-info.css` (or be applied directly on `.warning { cursor: pointer; }` there) rather than in the global stylesheet.

---

biodranik (commented):

---

PR #146, opened 2026-07-14
Feature/display routes

kiselev-dv:
* Display all routes going thru selected stop with variants
* Select route and variants for display
* Highlight OSM stops considered matched

### DIFF
--- a/src/services/ScheduleEncoding.ts
+++ b/src/services/ScheduleEncoding.ts
@@ -26,8 +26,10 @@ export interface ResolvedStopOnRoutePosition {
 
 export interface RouteSchedule {
   route: Route;
-  /** Index into sids[] — identifies the route direction at this stop. */
+  /** GTFS direction_id (0 or 1, 0 by default). */
   direction: number;
+  /** Index into sids[] — identifies the route direction at this stop (inferred from pos). */
+  directionSid?: number;
   positions: ResolvedStopOnRoutePosition[];
   tripTimes: TripTimes;
 }
@@ -64,7 +66,12 @@ export function decodeScheduleOnDate(
         if (!activeServices.has(entry.periods[i])) continue;
 
         const decodedPos = decodePosArray(allPos[entry.pos[i]]);
-        const direction =
+        const direction = entry.dir !== undefined ? entry.dir : (
+          decodedPos.nextStopId !== -1 && decodedPos.nextStopId !== currentSidIdx
+            ? decodedPos.nextStopId
+            : decodedPos.prevStopId
+        );
+        const directionSid =
           decodedPos.nextStopId !== -1 && decodedPos.nextStopId !== currentSidIdx
             ? decodedPos.nextStopId
             : decodedPos.prevStopId;
@@ -72,7 +79,7 @@ export function decodeScheduleOnDate(
         const key = `${route.routeId}=${direction}`;
         let rs = routeMap.get(key);
         if (!rs) {
-          rs = { route, direction, positions: [], tripTimes: { pos: [], tripId: [], arrivalTime: [], departureTime: [] } };
+          rs = { route, direction, directionSid, positions: [], tripTimes: { pos: [], tripId: [], arrivalTime: [], departureTime: [] } };
           routeMap.set(key, rs);
         }
 
--- a/src/services/ScheduleEncoding.ts
+++ b/src/services/ScheduleEncoding.ts
@@ -26,10 +26,10 @@ export interface ResolvedStopOnRoutePosition {
 
 export interface RouteSchedule {
   route: Route;
-  /** GTFS direction_id (0 or 1, 0 by default). */
-  direction: number;
-  /** Index into sids[] — identifies the route direction at this stop (inferred from pos). */
-  directionSid?: number;
+  /** GTFS direction_id (0 or 1) for the route. */
+  routeDirection?: number;
+  /** Index into sids[] (stop ids array) pointing to the next (previous for the last stop) stop */
+  adjacentStopSid?: number;
   positions: ResolvedStopOnRoutePosition[];
   tripTimes: TripTimes;
 }
@@ -66,16 +66,17 @@ export function decodeScheduleOnDate(
         if (!activeServices.has(entry.periods[i])) continue;
 
         const decodedPos = decodePosArray(allPos[entry.pos[i]]);
-        const directionSid =
+        const adjacentStopSid =
           decodedPos.nextStopId !== -1 && decodedPos.nextStopId !== currentSidIdx
             ? decodedPos.nextStopId
             : decodedPos.prevStopId;
-        const direction = entry.dir !== undefined ? entry.dir : directionSid;
+            
+        const routeDirection = entry.dir;
 
-        const key = `${route.routeId}=${directionSid}`;
+        const key = `${route.routeId}=${adjacentStopSid}`;
         let rs = routeMap.get(key);
         if (!rs) {
-          rs = { route, direction, directionSid, positions: [], tripTimes: { pos: [], tripId: [], arrivalTime: [], departureTime: [] } };
+          rs = { route, adjacentStopSid, routeDirection, positions: [], tripTimes: { pos: [], tripId: [], arrivalTime: [], departureTime: [] } };
           routeMap.set(key, rs);
         }
 
@@ -450,7 +451,7 @@ export function dateAsNumber(date: Date) {
 /**
  * Converts a day-number (days since 2000-01-01) to a yyMMdd number.
  * Uses UTC getters because EPOCH_MS is UTC midnight — using local getters would
- * shift the calendar date by one day in timezones behind UTC.
+ * shift the calendar date by one day in timezones behind UTC (e.g. America/Halifax).
  */
 function dayNumberAsYyMMdd(dayNumber: number): number {
   const d = new Date(EPOCH_MS + dayNumber * MS_PER_DAY);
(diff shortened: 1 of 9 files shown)


src/services/ScheduleEncoding.ts
```diff
@@ -64,15 +66,20 @@ export function decodeScheduleOnDate(
...
             ? decodedPos.nextStopId
             : decodedPos.prevStopId;
 
         const key = `${route.routeId}=${direction}`;
         let rs = routeMap.get(key);
         if (!rs) {
-          rs = { route, direction, positions: [], tripTimes: { pos: [], tripId: [], arrivalTime: [], departureTime: [] } };
+          rs = { route, direction, directionSid, positions: [], tripTimes: { pos: [], tripId: [], arrivalTime: [], departureTime: [] } };
           routeMap.set(key, rs);
         }
```
biodranik: `directionSid` is computed but never consumed anywhere (`schedule-preview.tsx` only destructures `route, direction, positions, tripTimes`), while the grouping `key` still uses the coarser `direction` (now `entry.dir`, i.e. 0/1 when the feed sets it). This can merge trips from different physical route branches that share the same GTFS `direction_id` into a single `RouteSchedule`; `schedule-preview.tsx` then shows `positions[0].lastStopName` as *the* destination for the whole merged group, so the "to X" label can be wrong for trips whose branch differs from `positions[0]`.

Suggest keying by `directionSid` (the previous per-branch value) and keeping `direction` purely for display. Also `direction`'s fallback and `directionSid` recompute the exact same ternary — compute once and reuse.

Separately, the new doc comment on `direction` ("0 or 1, 0 by default") is inaccurate: when `entry.dir` is absent, `direction` is actually the sid-based fallback value, not 0.

---

src/uielements/route-list.tsx
```diff
@@ -0,0 +1,238 @@
...
+                }
+            } catch (e) {
+                console.error('Failed to load routes', e);
+            } finally {
+                if (!cancelled) setLoading(false);
+            }
+        })();
+
+        return () => { cancelled = true; };
+    }, [routeIndex, routeIdsKey, gtfsStopIdsKey, reportRegion]);
```
biodranik, line 204: When `routeIds` becomes empty (e.g. the selection moves to a stop with no routes) this effect returns early without resetting `routesWithVariants` / `selectedRouteId` / `selectedVariantInx`, so the previous stop's route pills and map lines keep showing under the new selection. Reset the state (or guard the render) when `routeIds.length === 0`.

---

src/uielements/selection-info.tsx
```diff
@@ -146,26 +163,16 @@ function MatchInfo({ datasetName, properties, geometry, idTags }: MatchInfoProps
...
 
         <div className={"edit-actions"}>
             <AddOsmStopController id={properties.gtfsStopId} code={properties.gtfsStopCode} routeTypes={routeTypes} {...{ name, idTags }} />
         </div>
 
         {loading && <div>Loading OSM data...</div>}
 
-        <OsmElements setLoading={setLoading} osmFeatures={osmFeatures} tagActions={tagActions} parentLonLat={[lon, lat]} />
+        <OsmElements setLoading={setLoading} properties={properties} idTags={idTags} parentLonLat={[lon, lat]}
+            matched={datasetName ? !['nom', 'nos'].includes(datasetName) : false} />
```
biodranik: `matched` is recomputed here with the same expression already computed above at the `matched` const (used for `MatchArrowLayer`). Reuse that const instead of duplicating the ternary.

---

src/uielements/routes.tsx
```diff
@@ -8,10 +8,16 @@ export type RouteDisplayEntry = {
...
 }
 
+export type FullRouteDisplayEntry = {
+    routeKey: string;
+    coordinates: [number, number][];
+}
+
 type RoutesMapProps = {
-    entries: RouteDisplayEntry[];
+    entries?: RouteDisplayEntry[];
```
biodranik: After this PR, nothing passes `entries` to `RoutesMap` anymore — `selection-info.tsx` now renders `RouteList`/`fullRoutes` instead of the old `routesDisplayEntries`. The `entries`/`RouteDisplayEntry` prop and the next/prev-stop line-building code below (~L31-73) are now dead code; worth removing rather than keeping unused.

---

src/uielements/selection-info.tsx
```diff
@@ -44,61 +55,59 @@ export function SelectionInfo({ selection }: SelectionInfoProps) {
...
-    const osmFeatures = useMemo(() => parseJsonSafe(properties['osmFeatures'], []), [properties]);
+    const gtfsStopIds = useMemo(() => {
+        if (gtfsFeatures.length > 1) {
+            return gtfsFeatures.map((f: any) => f.id).filter(Boolean);
+        }
+        return properties.gtfsStopId ? [properties.gtfsStopId] : [];
+    }, [gtfsFeatures, properties.gtfsStopId]);
+    
     const routes = useMemo(() => parseJsonSafe(properties['gtfsRoutes'], null), [properties]);
+    const routeIdList = useMemo(() => Object.keys(routes || {}), [routes]);
```
biodranik: `routeIdList` now only reads the top-level `properties.gtfsRoutes`. The code this replaced handled clustered stops by falling back to `gtfsFeatures[].gtfsRoutes` (see the removed comment "For clusters routes are a property of gtfs features"). Without that fallback, `RouteList` receives `routeIds=[]` for cluster match types (sep/clu/mto/hub), so the new routes/variants panel never appears for them even though `gtfsLi` below still lists per-feature route IDs.

---

biodranik (commented):

Reviewed the diff and its callers (schedule-preview.tsx, selection-info.tsx, route-list.tsx, routes.tsx) across correctness, simplification, and dead-code angles. Found two confirmed correctness regressions (route-grouping key merges distinct branches; cluster stops lose the new routes panel), one stale-state bug in RouteList, and two simplification opportunities (duplicate computation, now-dead `RoutesMap.entries` code path). Details inline.

---

src/uielements/route-list.tsx
```diff
@@ -0,0 +1,264 @@
...
+                    let arr = grouped.get(t);
+                    if (!arr) { arr = []; grouped.set(t, arr); }
+                    arr.push(rwv);
+                }
+
+                if (grouped.size > 1) {
+                    return <div>
+                        {[...grouped.entries()].map(([type, routes]) => <div key={type} className="route-type-group">
+                            <span className="route-type-group-header"><b>{type}:</b> </span>
+                            {routes.map(r => <RoutePill {...routePillProps(r)} />)}
```
biodranik: `key` is now inside the spread object returned by `routePillProps` (L222), and both call sites render `<RoutePill {...routePillProps(r)} />` (here and L259) with no literal `key` attribute. This project uses the automatic JSX runtime (`@preact/preset-vite`, tsconfig `react-jsx` + `jsxImportSource: preact`), where `key` is only recognized as a literal JSX attribute — Preact's `jsx()` doesn't lift `key` out of spread props (it only extracts `ref`). So every `RoutePill` ends up with `key={undefined}` (a regression from the previous explicit `key={r.routeId}`), and `routeId` leaks in as a normal prop.

Pass it explicitly, e.g. `<RoutePill key={r.index.routeId} {...routePillProps(r)} />` at both call sites (and drop `key` from the returned object).

---

src/services/ScheduleEncoding.ts
```diff
@@ -64,15 +66,16 @@ export function decodeScheduleOnDate(
         if (!activeServices.has(entry.periods[i])) continue;
 
         const decodedPos = decodePosArray(allPos[entry.pos[i]]);
-        const direction =
+        const directionSid =
           decodedPos.nextStopId !== -1 && decodedPos.nextStopId !== currentSidIdx
             ? decodedPos.nextStopId
             : decodedPos.prevStopId;
+        const direction = entry.dir !== undefined ? entry.dir : directionSid;
```
biodranik: Keying by `directionSid` now looks correct. Small leftover: the doc comment on the `direction` field (interface, L29) still says "GTFS direction_id (0 or 1, 0 by default)". With this fallback, when `entry.dir` is absent `direction` is `directionSid` (a `sids[]` index), not 0 — worth updating the comment to match.

---

src/uielements/selection-info.css
```diff
@@ -3,6 +3,31 @@
     flex: 1;
     min-height: 0;
     border-top: 1px solid #ddd;
+    position: relative;
+}
+
+.close-button {
+    position: absolute;
```
biodranik, line 10: Minor UX: `.close-button` is `position: absolute` inside `#selection-info`, which is the `overflow-y: auto` scroll container. The button is anchored to the content box, so it scrolls out of view as the user scrolls the panel. If it should stay reachable, `position: sticky` (or rendering it as a sibling outside the scroll container) would keep it pinned.

---

biodranik (commented):

Re-reviewed after the "Resolve issues from review" commit. All five earlier findings are addressed: grouping now keys by `directionSid`, `routeIdList` falls back to per-feature `gtfsRoutes` for clusters, the empty-`routeIds` effect resets stale state, the duplicate `matched` computation is gone, and the dead `RoutesMap.entries` path is removed. 👍

Reviewed the new changes (route-type grouping, close button). Three points inline:
- `RoutePill` keys: `key` is now inside the spread props object, so under the automatic JSX runtime it isn't applied — regression from the previous explicit `key`. (route-list.tsx)
- Stale doc comment on `direction` ("0 by default"). (ScheduleEncoding.ts)
- Minor UX: the close button scrolls out of view with the panel content. (selection-info.css)

---

src/uielements/selection-info.css
kiselev-dv: Sticky looks worse, and since tabs aren't sticky there isn't much sense to fix this minor issue.

PR #150, opened 2026-07-27
Fix: Route ids for matches now contains direction indicator

kiselev-dv:
* Strip direction indicator
* Display route ids if routes failed to load or failed to find route by id
* Add loading animation for route pills

### DIFF
--- a/src/uielements/route-list.css
+++ b/src/uielements/route-list.css
@@ -14,6 +14,31 @@
     background: #e0e0e0;
 }
 
+.route-pill--skeleton {
+    opacity: 0.75;
+    cursor: default;
+}
+
+.route-pill--loading {
+    cursor: default;
+
+    background: linear-gradient(90deg, #e1e1e1, #5b5b5b);
+    background-size: 200% 200%;
+    animation: gradient-shift 4s linear infinite;
+}
+
+@keyframes gradient-shift {
+    0% {
+        background-position: 0% 50%;
+    }
+    50% {
+        background-position: 100% 50%;
+    }
+    100% {
+        background-position: 0% 50%;
+    }
+}
+
 .route-variant {
     cursor: pointer;
 }
--- a/src/uielements/route-list.tsx
+++ b/src/uielements/route-list.tsx
@@ -116,6 +116,7 @@ export function RouteList({ reportRegion, routeIds, routeTypes, gtfsStopIds }: R
     const [routeIndex, setRouteIndex] = useState<RouteIndexEntry[]>([]);
     const [routesWithVariants, setRoutesWithVariants] = useState<RouteWithVariants[]>([]);
     const [loading, setLoading] = useState(false);
+    const [indexLoading, setIndexLoading] = useState(false);
     const [selectedRouteId, setSelectedRouteId] = useState<string | null>(null);
     const [selectedVariantInx, setSelectedVariantInx] = useState<number | null>(null);
 
@@ -128,9 +129,13 @@ export function RouteList({ reportRegion, routeIds, routeTypes, gtfsStopIds }: R
         if (!reportRegion) return;
 
         let cancelled = false;
+        setIndexLoading(true);
 
         getRouteIndex(reportRegion).then(index => {
-            if (!cancelled) setRouteIndex(index);
+            if (!cancelled) {
+                setRouteIndex(index);
+                setIndexLoading(false);
+            }
         });
 
         return () => { cancelled = true; };
@@ -165,6 +170,7 @@ export function RouteList({ reportRegion, routeIds, routeTypes, gtfsStopIds }: R
             setRoutesWithVariants([]);
             setSelectedRouteId(null);
             setSelectedVariantInx(null);
+            setLoading(false);
             return;
         }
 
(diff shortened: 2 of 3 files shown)


src/uielements/route-list.tsx
```diff
@@ -203,12 +203,18 @@ export function RouteList({ reportRegion, routeIds, routeTypes, gtfsStopIds }: R
         return () => { cancelled = true; };
     }, [routeIndex, routeIdsKey, gtfsStopIdsKey, reportRegion]);
 
-    if (loading) {
-        return <div>Loading routes...</div>;
-    }
+    if (routesWithVariants.length === 0) {
+        if (!loading && routeIds.length === 0) return null;
```
biodranik: **Regression:** `Gtfs route types` is no longer rendered when a stop has no route ids. `nom`/`nos` selections typically carry `gtfsRouteTypes` but no `gtfsRoutes`, so `routeIdList` is empty, this returns `null`, and the types line disappears (the `&& !routeTypes` part of the old guard kept it).

```suggestion
        if (!loading && routeIds.length === 0 && !routeTypes) return null;
```

Two related points:
- With `routeIds` empty and `routeTypes` set, the block below still renders a bare `<b>Routes: </b>` with zero pills — worth wrapping that `<div>` in `routeIds.length > 0 &&`.
- `loading` can be stale-true here: the early return at L164-169 doesn't clear it, so switching selection while a variants fetch is in flight leaves `cancelled === true`, `finally { if (!cancelled) setLoading(false) }` never runs, and `loading` stays `true` for the rest of the session. Adding `setLoading(false)` to that early return fixes it.

---

src/uielements/route-list.tsx
```diff
@@ -203,12 +203,18 @@ export function RouteList({ reportRegion, routeIds, routeTypes, gtfsStopIds }: R
...
 
-    if (routesWithVariants.length === 0 && !routeTypes) {
-        return null;
+        const pillClass = cls('route-pill--skeleton', loading && "route-pill--loading");
```
biodranik: `loading` only covers the route-*variants* fetch. The region index fetch (L132 `getRouteIndex`) isn't tracked, and while it's in flight `routeIndex.length === 0` makes the effect at L163 early-return without touching `loading`. So on the **first selection in a region** the pills render in the plain `--skeleton` state — which in this PR means "couldn't resolve" — for the whole `routes.ndjson` download, then flip to `--loading`. The two states read identically to the user, just inverted in time.

An explicit flag around `getRouteIndex()` (set before, clear in `.then`/`.catch`) OR-ed into `pillClass` makes "still loading" honest, and keeps the static skeleton meaning "failed" (including the case where the index fetch rejects).

Nit: `className={cls('route-pill', 'route-pill--skeleton', loading && 'route-pill--loading')}` inline on L215 drops `pillClass` and the template string, matching L90.

---

src/uielements/route-list.tsx
```diff
@@ -203,12 +203,18 @@ export function RouteList({ reportRegion, routeIds, routeTypes, gtfsStopIds }: R
...
+        return <div>
+            {(routeTypes?.length || 0) > 0 &&
+                <div>Gtfs route types: <b>{routeTypes}</b></div>
+            }
+            <div><b>Routes: </b>
+                {routeIds.map(id => <span key={id} className={`route-pill ${pillClass}`}>{id}</span>)}
+            </div>
+        </div>;
```
biodranik, line 234: **Simplification:** this branch duplicates the route-types line (L223-225) and the `Routes:` wrapper (L263-265); the only real difference is which pills go inside. A single return keeps the two paths from drifting:

```tsx
const resolved = routesWithVariants.length > 0;
if (!resolved && !loading && routeIds.length === 0 && !routeTypes) return null;

return <div>
    {resolved && <RoutesMap fullRoutes={fullRouteEntries} />}
    {routeTypes && <div>Gtfs route types: <b>{routeTypes}</b></div>}
    {routeIds.length > 0 && (resolved ? renderResolvedPills() : renderSkeletonPills())}
</div>;
```

(`routeTypes` is `string | undefined`, so `{routeTypes && …}` replaces `(routeTypes?.length || 0) > 0` in both places.)

---

src/uielements/route-list.tsx
```diff
@@ -203,12 +203,18 @@ export function RouteList({ reportRegion, routeIds, routeTypes, gtfsStopIds }: R
...
 
-    if (routesWithVariants.length === 0 && !routeTypes) {
-        return null;
+        const pillClass = cls('route-pill--skeleton', loading && "route-pill--loading");
+        return <div>
+            {(routeTypes?.length || 0) > 0 &&
+                <div>Gtfs route types: <b>{routeTypes}</b></div>
+            }
+            <div><b>Routes: </b>
+                {routeIds.map(id => <span key={id} className={`route-pill ${pillClass}`}>{id}</span>)}
```
biodranik: **Incomplete vs. the stated goal** ("display route ids if we failed to find route by id"): this fallback is gated on *nothing* resolving. For ids `A,B` where `A` is found in the index and `B` isn't (or `B` has no variant passing through `gtfsStopIds`), the main path renders only `A` and `B` vanishes silently — the partial case is the likely one in practice.

Rendering unresolved ids as skeleton pills alongside the resolved ones covers it, and pairs with a cheap perf fix: L177 `routeIndex.filter(e => routeIds.includes(e.routeId))` is O(index × ids) over the entire region index on every selection. A `useMemo`'d `Map<routeId, RouteIndexEntry>` built once per `routeIndex` makes the lookup O(ids) *and* directly gives you the unresolved set.

---

src/uielements/selection-info.tsx
```diff
@@ -98,11 +98,12 @@ function MatchInfo({ datasetName, properties, geometry, idTags, reportRegion }:
     
     const routes = useMemo(() => parseJsonSafe(properties['gtfsRoutes'], null), [properties]);
     const routeIdList = useMemo(() => {
-        if (routes) return Object.keys(routes);
+        const stripDir = (id: string) => id.replace(/ d\d+$/, '');
+        if (routes) return [...new Set(Object.keys(routes).map(stripDir))];
```
biodranik, line 102: The reasoning here is not obvious from the code and deserves one line: ` d<N>` is the GTFS `direction_id` that the report appends to route ids, while `routes.ndjson` is keyed by the bare `route_id` (direction lives on the variants as `dir`), so the suffix has to go before the lookup.

Two caveats:
- `route_id` is an opaque agency string. One that genuinely ends in ` d1` gets silently mangled and then fails to resolve, with no way to tell. Looking the id up as-is first and only falling back to the stripped form would be lossless — and would let the fallback pills in `RouteList` show what the data actually contains.
- L135 still prints the raw keys, so a multi-feature selection shows `R1 d0 R1 d1` in the per-feature list next to a single deduped `R1` pill from `RouteList`. Applying the same strip+dedup there (shared helper) keeps the two consistent.

---

src/uielements/route-list.css
```diff
@@ -14,6 +14,31 @@
...
+    0% {
+        background-position: 0% 50%;
+    }
+    50% {
+        background-position: 100% 50%;
+    }
+    100% {
+        background-position: 0% 50%;
+    }
+}
```
biodranik, line 40: Minor:
- The gradient ends at `#5b5b5b`; with the default dark text the route id is barely readable at the dark end of the sweep. A lighter range (e.g. `#e1e1e1 → #b8b8b8`) keeps the shimmer visible without hurting contrast.
- `cursor: default` is redundant — `--loading` is only ever applied together with `--skeleton` (route-list.tsx L209), which already sets it.
- `gradient-shift` is a global keyframe name in a global stylesheet; something like `route-pill-shimmer` avoids collisions.
- Worth guarding: `@media (prefers-reduced-motion: reduce) { .route-pill--loading { animation: none; } }`.

---

biodranik (commented):

Stripping the direction suffix is the right fix for the lookup, and showing the raw ids instead of a bare "Loading routes..." is a clear improvement. A few things found in the new render path:

**Confirmed issues**
1. `Gtfs route types` is dropped for stops with no route ids (`nom`/`nos` selections) — the old guard kept it, the new one returns `null`. (route-list.tsx L207)
2. The region-index fetch isn't part of `loading`, so the *first* selection in a region shows the "failed" skeleton for the whole `routes.ndjson` download. (L209)
3. Partial resolution failures still hide ids: the fallback only fires when nothing resolved, so an unresolvable id next to a resolvable one disappears silently. (L215)
4. `loading` is never cleared by the effect's early return, so it can stay `true` indefinitely after a selection switch during an in-flight fetch — which now leaks a bare `<b>Routes: </b>` header. (L164-169)

**Alternative worth considering**

The suffix is currently discarded. Since `routes.ndjson` variants already carry `dir` and the UI labels them ↑/↓, parsing `id` into `{ routeId, dir }` and using `dir` to filter/preselect the matching variant would preserve what the report now tells us — which directions actually serve this stop. Pros: no information loss, and it's robust against a `route_id` that legitimately ends in ` d1`. Cons: more state for something the `gtfsIds.includes(stopId)` filter mostly already narrows down, so today it buys little visible benefit — reasonable to keep stripping, just worth a comment recording the assumption.

Cleanest option overall is backend-side: keep `route_id` as the key and expose direction as a value field, which removes the frontend parsing entirely. Out of scope for this PR, but it's what makes the ` d\d+` regex unnecessary.

---

PR #151, opened 2026-07-28
Read index.tsv by column name, and facet matches by strategy

kiselev-dv:
`parseIndex` read columns positionally, so a report with a different column layout silently misread every column after the difference — a coordinate as a byte offset, and a bad range request instead of an error. Columns are resolved by header name now; `kind` and the older `status_detailed` are both accepted as the category column. A header this build cannot read is reported in the panel rather than drawn as a region with no stops, and the range requests check the response before parsing it: an unknown detail type asked for a file the report usually does not have, and `JSON.parse` then ran on the 404 body inside an un-awaited callback, so the panel never opened and nothing said why.

How a stop was matched — by id, routes, name, platform code, stop_area, or name against a conflicting id — is not a kind of stop, since several can hold at once. It shows as badges on the selected stop, read from the `strategies` object in the detail body, which names them. The alias table was silently dropping `platform` and `area`, so a stop matched only that way carried no badges at all; all seven are named now. The index's `strategies` bit field is left for a facet that is not built: nothing decodes it, so the one file loaded eagerly per region does no per-row work for it.

The category table had two copies, in `report.tsx` and `selection-info.tsx`, which had already drifted word by word across all twelve codes — one table now, beside the strategies it belongs with. `route-stops.ndjson` likewise had two identical fetchers behind two caches, so a route seen in both panels was fetched twice; one shared module, and a failed range request is evicted rather than cached, which used to pin a panel at "Loading…" for the rest of the session.

Preview hides the report's stops through the filter `StopsLayer` was built around instead of unmounting the layer, which dropped and re-ingested the whole source on every toggle. That also retires the saved-and-restored selection: while preview was on nothing could change it except the deep-link write, which the restore then threw away. Its own source is clustered, so a click below `clusterMaxZoom` handed back a cluster with no id — it zooms in now rather than opening a panel of `undefined`.

GTFS ids are free-form UTF-8 and do occur with a `#` or a `/`, either of which truncated the selection hash. It percent-encodes them, and decodes leniently so a link made before this still resolves.

`ScheduleEncoding.ts`, `schedule.types.d.ts` and `FeedMetaT` go with `schedule-preview.tsx`, their only consumer.

`tsc -b` and `vite build` pass.


src/services/matchIndex.ts
```diff
@@ -0,0 +1,149 @@
...
+        rows.push({
+            id: c[at['gtfs:id']],
+            status: at['status'] !== undefined ? c[at['status']] : '',
+            code,
+            type: c[at['type']],
+            lon: parseFloat(c[at['lon']]),
+            lat: parseFloat(c[at['lat']]),
+            byteStart: parseInt(c[at['byte_start']], 10),
+            byteEnd: parseInt(c[at['byte_end']], 10),
+            strategies: parseStrategies(at['strategies'] !== undefined ? c[at['strategies']] : undefined, code),
```
biodranik: `IndexRow.strategies` is never read: `grep -rn strategies src/` shows the only consumers are `STRATEGIES` and `strategiesForCode`, both used by `selection-info.tsx` from the *detail body*, not from the index. So `STRATEGY_BITS`, `parseStrategies` and this field are dead, and the multi-select facet the PR description promises ("Added as a multi-select filter under the categories… the header says so") does not exist in `report.tsx`.

Either land the facet, or drop the bit-field path and the field and fix the description — the index is the one file loaded eagerly per region, so per-row work that nothing consumes is the one place it costs.

---

src/services/matchIndex.ts
```diff
@@ -0,0 +1,149 @@
...
+
+/** What was true about a match. Several can hold for one stop. */
+export type Strategy = 'id' | 'routes' | 'name' | 'conflict' | 'narrowed';
+
+export const STRATEGIES: { key: Strategy; label: string; help: string }[] = [
+    { key: 'id', label: 'by id', help: 'Matched by GTFS id or code' },
+    { key: 'routes', label: 'by routes', help: 'Matched by routes going through the stop' },
+    { key: 'name', label: 'by name', help: 'Matched by name' },
+    { key: 'conflict', label: 'id conflict', help: 'Matched by name, but an id on the OSM feature disagrees' },
+    { key: 'narrowed', label: 'narrowed by routes', help: 'The routes cut the answer down to the features they serve' },
```
biodranik: Proofreading: "The routes cut the answer down to the features they serve" — "the answer" has no referent for a user, and "cut down" is colloquial. Suggest "Routes narrowed the candidates to the features they serve".

Also line 25: "but an id on the OSM feature disagrees" → "but the OSM feature carries a different GTFS id" reads more concretely.

---

src/uielements/report.tsx
```diff
@@ -26,6 +28,9 @@ type Category = {
 };
 
 const CATEGORIES: { [code: string]: Category } = {
+    mat: { group: 'matched', label: 'match', color: 'green', help: 'Matched to one or more OSM features. How it matched is the strategies column' },
```
biodranik: Two issues here.

1. Wording: a user tooltip should not name a file column. `selection-info.tsx` already says it better — "Matched to one or more OSM features; the badges say how."

2. Adding `mat` required editing two tables: `CATEGORIES` here and `MATCH_LABELS` in `selection-info.tsx:96`. They duplicate `label` + `help` for all 12 codes and have already drifted — e.g. `sep` here reads "successfuly separated" / "one or many GTFS", while `selection-info.tsx` has "successfully separated" / "one or many GTFS stops"; `nom`, `nos`, `clu`, `mto`, `hub` differ too.

Simplification: move the code table to `matchIndex.ts` next to `STRATEGIES` as `{ group, color, label, help }`, import it in both. `report.tsx` reads `group`/`color`, `selection-info.tsx` reads `label`/`help`, and the typos disappear with the copy.

---

src/uielements/report.tsx
```diff
@@ -196,6 +164,7 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
 
     const [rows, setRows] = useState<IndexRow[]>([]);
     const [selectedDatasets, updateSelectedDatasets] = useState<DatatsetsSelectonT>(defaultSets);
+    const [savedDatasets, setSavedDatasets] = useState<DatatsetsSelectonT | null>(null);
```
biodranik: `savedDatasets` restores a snapshot nothing can have changed: while preview is on, both checkboxes are `disabled` (L326, L335) and `stopsLayer` is unmounted (L365), so the state at toggle-off is exactly `{...saved, preview: true}`. The one writer that *can* still run is the deep-link effect at L264 (`updateSelectedDatasets(prev => ... {...prev, [row.code]: true})`) — and the restore then throws that write away. So the machinery is a no-op in the normal path and a small bug in the other one.

Simplification, ~14 lines out:

```js
onChange={e => updateSelectedDatasets({ ...selectedDatasets, preview: (e.target as HTMLInputElement).checked })}
```

and hide the stops through the filter that already exists rather than unmounting the layer:

```js
const selectedCodes = selectedDatasets['preview'] ? [] : CATEGORY_CODES.filter(c => selectedDatasets[c]);
const stopsLayer = rows.length > 0 && <StopsLayer ... />;
```

That also avoids tearing down and re-ingesting the whole GeoJSON source on every preview toggle — `StopsLayer` was built around `map.setFilter` for exactly this ("The source data is never rebuilt on toggle", L415).

---

src/uielements/report.tsx
```diff
@@ -222,10 +189,11 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
...
 
+    const featureCollection = useMemo(() => buildFeatureCollection(rows), [rows]);
+
     // Range-fetch a single stop's detail object and turn it into a selection.
     const selectStop = useCallback(async (loc: StopLocator, source: 'map-click' | 'url-hash') => {
-        const file = fileFor[loc.type];
-        if (!file) return;
+        const file = detailFileFor(loc.type);
 
         const res = await fetch(`${DATA_BASE_URL}/${reportRegion}/${file}`, {
```
biodranik, line 195: `detailFileFor` falling back to `report.ndjson` turned a silent no-op (`if (!file) return;`) into a request that is very likely a 404 — and nothing checks `res.ok`, so `JSON.parse` runs on the error body and throws inside an un-awaited async callback. The user sees a panel that never opens and no message; only the console has a rejected promise.

This is the spot where failing loudly pays for itself, in dev and in production:

```js
const res = await fetch(url, { headers: { Range: `bytes=${loc.byteStart}-${loc.byteEnd}` } });
if (!res.ok) throw new Error(`${res.status} for ${file} ${loc.byteStart}-${loc.byteEnd}`);
```

Same for the `ir-preview.ndjson` load at L233 (no `.ok`, no `.catch`) and for `parseIndex` returning `[]` on a missing column — a `console.error` plus an empty map looks identical to a region with no stops. Surfacing "index.tsv could not be read" in the panel would make a bad report obvious immediately.

---

src/uielements/report.tsx
```diff
@@ -259,12 +227,22 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
         updateSelection({ feature, datasetName: 'preview', reportRegion, idTags }, 'map-click');
```
biodranik, line 247: The preview source is clustered (`cluster: true, clusterMaxZoom: 10`, L548-551), so at zoom ≤ 10 a click hands this a *cluster* feature — `{cluster: true, cluster_id, point_count}`, no `id`/`lat`/`lon`. That object is truthy, so `Preview`'s `if (!stop)` guard passes and the panel renders "GTFS ID: undefined", "Position: ," and "No routes", plus a `#…/preview/undefined` hash.

```js
if (!feature || feature.properties?.cluster) return;
```

or, better for the user, zoom to the cluster instead of opening the panel.

---

src/uielements/report.tsx
```diff
@@ -364,13 +346,23 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
...
+                        if (checked) {
+                            setSavedDatasets({ ...selectedDatasets });
+                            updateSelectedDatasets({ ...selectedDatasets, preview: true });
+                        } else {
+                            const restored = { ...(savedDatasets || defaultSets), preview: false };
+                            setSavedDatasets(null);
+                            updateSelectedDatasets(restored);
+                        }
+                    }} />
                 <span className={'match-dataset'} title={'Preview timetables for all matched'}>Preview</span>
```
biodranik: Stale after `SchedulePreview` was replaced — the panel no longer shows timetables. Suggest `title={'Show the matcher\'s stop positions and routes'}` (or whatever wording fits) and `Preview` stays as the label.

Two more places say the same outdated thing:
- L520-521: `// Used only for the preview overlay (preview.geojson), which is unchanged by the index/NDJSON migration.` — the file is now `ir-preview.ndjson`, and the sentence describes a migration rather than the code. Something like `// The preview overlay: one clustered symbol layer over ir-preview.ndjson.`
- `routing.ts:41`: `// …/preview/{id} for timetable preview, …`

---

src/uielements/preview.tsx
```diff
@@ -0,0 +1,311 @@
...
+            .then(r => r.text())
+            .then(text => JSON.parse(text.trim()).routes as IrPreviewRoute[])
+            .catch(() => []);
+    }
+    return stopRoutesCache[key];
+}
+
+const routeVariantCache: { [key: string]: Promise<RouteVariant[]> } = {};
+
+function getRouteVariants(reportRegion: string, route: IrPreviewRoute): Promise<RouteVariant[]> {
```
biodranik: This function, `RouteVariant`, `routeVariantCache` and the cache key `${region}:${routeId}` are a verbatim copy of `route-list.tsx:19-71`, over the same `route-stops.ndjson` byte ranges. Two module-level caches for identical data means the same route is fetched twice if a user sees it in both panels.

Export one `getRouteVariants(region, routeId, byteOffset, byteLength)` (plus `RouteVariant`) from a shared module — `route-list.tsx`'s `RouteIndexEntry` and `IrPreviewRoute` only differ in the field names holding those two numbers.

Separately, there is no `.catch` here: a failed range request leaves `loadingVariants` true forever ("Loading route variants…"), raises an unhandled rejection, and caches the rejected promise so reselecting the route never recovers. `route-list.tsx` has the same gap, so fixing it in the shared copy fixes both — reject-and-evict is enough:

```js
.catch(e => { delete routeVariantCache[key]; throw e; })
```

with a `.catch` on the caller that clears the loading flag.

---

src/uielements/preview.tsx
```diff
@@ -0,0 +1,311 @@
...
+                        The station is what the map labels; this platform is where the passenger stands.
+                    </div>
+                </div>
+            )}
+
+            <div>
+                <h5>Routes</h5>
+                {!hasRoutes ? (
+                    <div>No routes</div>
+                ) : routes.length === 0 ? (
```
biodranik: `routes.length === 0` doubles as the loading flag, so a stop whose routes come back empty shows "Loading routes…" forever. That includes the failure path: `getStopRoutes` swallows every error into `[]` (L63) and caches it, so one transient 404/416 pins the stop at "Loading routes…" for the rest of the session with nothing in the console.

`loadingVariants` right below already has the right shape — mirror it:

```js
const [loadingRoutes, setLoadingRoutes] = useState(false);
...
{!hasRoutes ? <div>No routes</div>
 : loadingRoutes ? <div>Loading routes&hellip;</div>
 : routes.length === 0 ? <div>Could not load routes</div>
 : ...}
```

and in `getStopRoutes`, evict on failure (`.catch(e => { delete stopRoutesCache[key]; throw e; })`) so a retry is possible.

---

src/uielements/preview.tsx
```diff
@@ -0,0 +1,311 @@
...
+    const lonlat = (selection?.feature.geometry as { coordinates: number[] } & any)?.coordinates;
+    const { onReportSelect } = useContext(SelectionContext);
+    const reportRegion = (selection as any)?.reportRegion;
+
+    const [routes, setRoutes] = useState<IrPreviewRoute[]>([]);
+    const [selectedRouteId, setSelectedRouteId] = useState<string | null>(null);
+    const [variants, setVariants] = useState<RouteVariant[]>([]);
+    const [loadingVariants, setLoadingVariants] = useState(false);
+
+    // stringifyProperties turns numbers into strings on the way through the map layer.
```
biodranik: This is inaccurate: `stringifyProperties` (`report.tsx:611`) only JSON-stringifies arrays and objects — numbers pass through untouched, and MapLibre keeps numeric GeoJSON properties numeric. So the `Number(...)` conversions here and on L147 guard against nothing, and the comment sends the next reader looking for a conversion that doesn't happen.

Either drop them —

```js
const hasRoutes = stop?.routesStart != null;
```

— or, if a report is actually known to write these as strings, say that instead. (`asList` on L109 is the genuine case and its comment is correct.)

---

src/uielements/preview.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+    if (import.meta.env.DEV) {
+        console.log('Preview', selection);
+    }
+
+    if (!stop) {
+        return <div id="selection-info">No stop selected</div>;
+    }
+
+    const fullRouteEntries: FullRouteDisplayEntry[] = variants.map((v, i) => {
```
biodranik: `fullRouteEntries` gets a new identity on every render, and `RoutesMap`'s effect depends on it (`routes.tsx:95`) — its cleanup wipes the `routes` source to an empty collection and the effect re-`setData`s it, so the drawn line is cleared and re-uploaded on every re-render of this panel. `route-list.tsx:141` wraps the same computation in `useMemo` for that reason; do the same here, keyed on `[variants, routes, selectedRouteId]`.

While moving it: `const route = routes.find(...)` on L190 is loop-invariant — hoist it above the `map`.

---

src/uielements/preview.tsx
```diff
@@ -0,0 +1,311 @@
...
+    }
+    return routeVariantCache[key];
+}
+
+/**
+ * Why a stop kept the position its feed gave it. The matcher's own vocabulary, spelled out —
+ * a reviewer looking at a stop in the wrong place needs to know whether OSM was missing, the
+ * match was refused, or the match was taken and then rejected as too far.
+ */
+const NO_ANCHOR_REASON: { [code: string]: string } = {
```
biodranik, line 73: Proofreading these — they read as the sentence tail after "Not anchored — ", which works, but two are opaque:

- `notANode`: "which has no position the map holds" → "which has no single point to place the stop at".
- `tooFar`: "beyond the shift the anchoring allows" — "the shift" is undefined for the reader → "farther away than the anchoring is allowed to move the stop".
- `severalPositions` / `idConflict` / `noPosition` read well as they are.

Also L204: `stopLink` interpolates a raw GTFS id into the hash. GTFS ids are free-form UTF-8 and do occur with spaces or `#`/`/`; a `#` truncates the link and `parseSelectionHash`'s `[^/]+` cuts at a slash. `encodeURIComponent(id)` here plus `decodeURIComponent` in `parseSelectionHash` covers it (`app.tsx:142` writes the hash unencoded too, so the fix belongs in both).

---

biodranik (commented):
Reviewed the column-by-name parse, the new `Preview` panel and the strategy badges, plus the impacted callers (`report.tsx`, `selection-info.tsx`, `route-list.tsx`, `routes.tsx`, `routing.ts`, `app.tsx`). Inline comments carry the specifics; two things that have no line to sit on:

**Dead modules left by removing `schedule-preview.tsx`.** It was the only consumer of `services/ScheduleEncoding.ts` (18 KB), `services/schedule.types.d.ts` (7.9 KB) and `FeedMetaT` in `types.ts` — `grep -rn "ScheduleEncoding\|schedule.types\|FeedMetaT" src/` now returns only their own definitions and each other. Vite tree-shakes them out of the bundle, so this is source weight, not payload, but it should go with the component. (`Report.liveUpdates` and `report-table.tsx`'s use of it are unrelated and still live.)

**Alternatives worth a thought, briefly:**

1. *Strategies as a bit field vs. a name list.* The bit field is the compact choice for the eagerly-loaded index, but 1/2/4/8/16 live only in `STRATEGY_BITS` with nothing tying them to the generator — renumber on that side and the UI silently mislabels rather than fails. `id,routes,name` costs ~10 bytes a row, is self-describing, needs no shared constant, and gzips to roughly the same size on the wire. Given nothing reads the column yet, this is a cheap moment to pick. Pro of bits: fixed width, trivial `&` filtering if the facet lands.

2. *Format detection.* `CATEGORY_COLUMNS` probing `kind` then `status_detailed` reads whichever it finds, so a report that is half-migrated is half-read. `matchMeta.matcherVersion` is already fetched — branching on it makes the two formats explicit and lets the loader reject one it does not know. Cost: the generator has to bump it reliably.

3. *Preview/report exclusivity.* It is currently enforced in three places — disabled checkboxes, an unmounted stops layer, and the saved/restored selection. One `mode: 'report' | 'preview'` with `selectedCodes` derived from it collapses all three (see the `savedDatasets` comment).

---

src/uielements/selection-info.tsx
```diff
@@ -154,11 +181,19 @@ function MatchInfo({ datasetName, properties, geometry, idTags, reportRegion }:
...
         <h2>{name}{matchInfo && <span className="match-type-tag" title={matchInfo.help}>{matchInfo.label}</span>}</h2>
 
+        {strategies.length > 0 && <div className="match-strategies">
+            {strategies.map(key => {
+                const s = STRATEGIES.find(s => s.key === key)!;
+                return <span className="match-type-tag" key={key} title={s.help}>{s.label}</span>;
+            })}
+        </div>}
+
         <DatasetHelp datasetName={datasetName} />
```
biodranik, line 190: `DatasetHelp` (L577-610) has no `mat` branch, so with this PR every matched stop in a current report renders its placeholder: *"One day here will be info text for mat"*. `datasetName` is `loc.subcategory` → `IndexRow.code` → the `kind` column, which is `mat` for all matched stops now.

It is also the third copy of the category text the PR just consolidated, and it has already drifted: it folds `nic` into `mnm` ("Matched OSM stops by GTFS stop Name"), while `CATEGORIES.nic` says "matched by name but mismatched by id".

One source of truth fixes both:

```js
function DatasetHelp({ datasetName }: { datasetName?: string }) {
    const help = datasetName ? CATEGORIES[datasetName]?.help : undefined;
    return help ? <p><i>{help}</i></p> : null;
}
```

(or drop the component — the `.match-type-tag` on L188 already carries the same text as its `title`.)

The comment on L576, `// datasetName is the index.tsv status_detailed 3-letter code`, is stale either way: `matchIndex.ts` reads the code from `kind` first and `status_detailed` only for older reports.

---

src/uielements/report.tsx
```diff
@@ -154,7 +93,7 @@ function buildFeatureCollection(rows: IndexRow[]): GeojsonDataT {
             geometry: { type: 'Point', coordinates: [r.lon, r.lat] },
             properties: {
                 gtfsStopId: r.id,
-                category: r.status === 'm' ? 'matched' : 'not-matched',
+                category: matchedGroup(r) ? 'matched' : 'not-matched',
```
biodranik: `category` is written here and read by nothing — `grep -rn category src/ | grep -v CATEGOR | grep -v subcategory` is empty; both the layer filter (L409) and the icon expression (L446) key off `subcategory`.

So `matchedGroup`, `IndexRow.status` and the optional `status` column in `parseIndex` exist only to compute a dead property. Dropping all four is ~14 lines out and one fewer optional column to reason about. The fallback branch is not load-bearing either: a code missing from `CATEGORIES` is already invisible on the map regardless of what `status` says, since it is never in `CATEGORY_CODES` and so never in `selectedCodes`.

---

src/uielements/report.tsx
```diff
@@ -206,14 +147,20 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
...
         }
+        setLoadError(null);
         fetch(`${DATA_BASE_URL}/${reportRegion}/index.tsv?t=${Date.now()}`)
-            .then(r => r.text())
-            .then(t => { if (!cancelled) setRows(parseIndex(t)); });
+            .then(r => {
+                if (!r.ok) throw new Error(`${r.status} for index.tsv`);
+                return r.text();
+            })
+            .then(t => { if (!cancelled) setRows(parseIndex(t)); })
```
biodranik: `parseIndex` validates the header but not the values, so the failure the PR set out to remove survives one level down: a row whose `kind` is not in `CATEGORIES` gets no preloaded icon (L470 iterates `CATEGORY_CODES`), no checkbox, no count, and can never enter `selectedCodes` — the stops just are not on the map, and the panel says nothing. That is exactly "an unreadable index and a region with no stops produce the same empty map", for a report one code newer than this build.

Cheap to close, in the same `loadError` channel:

```js
.then(t => {
    if (cancelled) return;
    const parsed = parseIndex(t);
    const unknown = [...new Set(parsed.map(r => r.code))].filter(c => !CATEGORIES[c]);
    if (unknown.length > 0) setLoadError(`index.tsv has categories this build does not know: ${unknown.join(', ')}`);
    setRows(parsed);
})
```

---

src/uielements/report.tsx
```diff
@@ -251,20 +204,52 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
             byteEnd: p.byteEnd,
             lon, lat,
             subcategory: p.subcategory,
-        }, 'map-click');
+        }, 'map-click').catch(e => {
+            console.error('Could not load stop detail', e);
+            setLoadError(`Could not load the stop: ${e.message}`);
```
biodranik: `loadError` is cleared in one place only — L150, on region change. A single failed click therefore leaves "Could not load the stop: 404 …" pinned above the report for the rest of the session, including after every subsequent successful selection, which reads as if the report itself is broken.

Clearing it where the attempt starts keeps the banner about the current state:

```js
setLoadError(null);
selectStop({ … }, 'map-click').catch(…)
```

Same for the deep-link path at L273 and the preview load at L232.

---

src/uielements/report.tsx
```diff
@@ -251,20 +204,52 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
...
+    }, [map, reportRegion, idTags, updateSelection]);
 
-    // Load preview.geojson lazily when the preview toggle is on.
+    // Load ir-preview.ndjson lazily when the preview toggle is on.
     useEffect(() => {
         if (selectedDatasets['preview'] && !previewData) {
-            fetch(`${DATA_BASE_URL}/${reportRegion}/preview.geojson`)
-                .then(r => r.json())
-                .then(setPreviewData);
+            fetch(`${DATA_BASE_URL}/${reportRegion}/ir-preview.ndjson`)
```
biodranik, line 255: No cancellation guard here, unlike the index load right above (L144). `ir-preview.ndjson` is a whole-region file, so it is the slowest fetch in the panel: switch region while it is in flight and the region effect nulls `previewData`, then this resolves and puts the previous region's stops back on the map — clustered, clickable, and deep-linkable under the new region's name.

The same `let cancelled = false` / `return () => { cancelled = true; }` as L144 covers it (guarding both `setPreviewData` and `setLoadError`).

---

src/uielements/selection-info.tsx
```diff
@@ -58,20 +60,45 @@ export function SelectionInfo({ selection }: SelectionInfoProps) {
...
+ * what was true about it.
+ */
+function strategiesOf(properties: { [key: string]: any }, datasetName?: string): Strategy[] {
+    const written = properties['strategies'];
+    if (written) {
+        const keys = typeof written === 'string' ? parseJsonSafe<any>(written, null) : written;
+        // Accepts either shape: a list of names, or an object keyed by name whose
+        // values carry each strategy's detail.
+        const names: string[] = Array.isArray(keys) ? keys : (keys ? Object.keys(keys) : []);
+        const known = names.map(n => STRATEGY_ALIASES[n]).filter(Boolean) as Strategy[];
```
biodranik: Simplification: `STRATEGY_ALIASES` maps six of seven keys to themselves, and its own doc comment names the hazard it creates — a second table to keep in sync with `STRATEGIES`, where forgetting a row drops badges silently (which is how `platform` and `area` went missing). Only `idConflict` is actually an alias:

```js
const alias = (n: string) => n === 'idConflict' ? 'conflict' : n;
const known = names.map(alias).filter(n => STRATEGIES.some(s => s.key === n)) as Strategy[];
```

Same result for every input the table handles, one table instead of two, and a new strategy is named once in `STRATEGIES`.

While here: `narrowed` is only appended inside `if (written)`, so a report that writes `narrowedByRoutes` without a `strategies` object loses the badge — worth hoisting the check out if that shape can occur.

---

src/services/matchIndex.ts
```diff
@@ -0,0 +1,168 @@
...
+
+export const CATEGORIES: { [code: string]: Category } = {
+    mat: { group: 'matched', label: 'match', color: 'green', help: 'Matched to one or more OSM features; the badges say how' },
+    // Older reports split a match across these four by strategy, one row each; current
+    // ones use `mat` and name the strategies separately. Kept so an old report still renders.
+    mid: { group: 'matched', label: 'match-id', color: 'green', help: 'Stop matched by GTFS id or code' },
+    mrt: { group: 'matched', label: 'match-routes', color: 'green', help: 'Stop matched by routes going through it' },
+    mnm: { group: 'matched', label: 'match-name', color: 'green', help: 'Stop matched by name' },
+    nic: { group: 'matched', label: 'name-id-conflict', color: 'green', help: 'Stop matched by name but mismatched by id' },
+    gen: { group: 'matched', label: 'match-generic', color: 'green', help: 'Matched to a stop without name or code nearby' },
```
biodranik: Proofreading: "Matched to a stop without name or code nearby" parses two ways — most readers land on "no name or code *nearby*". Suggest **"Matched to a nearby OSM stop that has neither a name nor a code"**.

Two more in this table:
- `nos`: "No OSM elements of matching transport mode found in the area" → "No OSM element of a compatible transport mode was found in the area" (`preview.tsx` already says "compatible mode", and a singular reads better for a per-stop verdict).
- `hub`: "Many OSM platforms or stops matched to one station by name" — `mto`/`clu`/`sep` all use the active "matched … by name"; this one is passive mid-sentence. → "Many OSM platforms or stops matched one station by name".

`STRATEGIES` reads well; `platform`'s `local_ref`/`ref` wording is the right OSM vocabulary.

---

src/uielements/preview.tsx
```diff
@@ -0,0 +1,338 @@
...
+    idConflict: 'every feature it matched carries another stop’s id',
+    severalPositions: 'the features it matched sit at different positions',
+    notANode: 'it matched only a way or a relation, which has no single point to place the stop at',
+    tooFar: 'the feature it matched is farther away than the anchoring is allowed to move the stop',
+    noPosition: 'the feed gives the stop no coordinate',
+    noOsmNearby: 'no OSM feature of a compatible mode was found anywhere near it',
+    noMatch: 'OSM features were found nearby, but none convinced any strategy',
+    genericOnly: 'the only thing nearby was an untagged stop, which is proximity rather than evidence',
```
biodranik: Two of these read imprecisely for an OSM audience:

- `genericOnly`: "an untagged stop" is not what the matcher means — a feature with no tags at all would never be recognised as a stop. It is a stop with the transport tags but no `name`/`ref`. → **"the only thing nearby was a stop with no name or code, which is proximity rather than evidence"** (and it then matches `CATEGORIES.gen`, which is the same verdict).
- `noMatch`: "none convinced any strategy" reads as if the strategies have opinions. → **"but none satisfied any matching strategy"**.

The rest, `tooFar` and `notANode` included, read well.

---

biodranik (commented):
Every thread from the previous round is addressed on `30bf3ae`; resolving them. Eight new items inline, plus these three that have no line in the diff to hang on:

**`routing.ts:6-22` — the `hashchange` listener is never removed.** `subscribe` adds an anonymous wrapper but the returned cleanup calls `removeEventListener("hashchange", callback)`, which was never registered, so nothing is ever detached. `subscribe` is also a fresh arrow on every render, so `useSyncExternalStore` re-subscribes and adds another wrapper each time. Every hash change then re-runs all of them. Pre-existing, but this PR is the one touching the file:

```js
export function useHash() {
    const subscribe = useCallback((callback: () => void) => {
        const onChange = () => { … callback(); };
        window.addEventListener("hashchange", onChange);
        return () => window.removeEventListener("hashchange", onChange);
    }, []);
    return useSyncExternalStore(subscribe, () => window.location.hash);
}
```

**Three copies of one caching idiom.** `routeVariantCache`, `routeIndexCache` and `stopRoutesCache` are now the same shape — a module-level `{[key]: Promise<T>}` plus `.catch(e => { delete cache[key]; throw e; })`. A four-line `memoFetch(cache, key, fn)` next to `getRouteVariants` collapses all three and puts the evict-on-failure rule, which is the part that is easy to forget, in one place.

**Preview styles live in `route-list.css`.** `.anchor-state*`, `.platform-code` and `.station-children` belong to `preview.tsx`, which imports `route-list.css` only to reach them. A `preview.css` beside the component would match the rest of the tree; `route-list.css` stays for `RouteList` and the route pills both panels share.

### Alternatives considered, and why the PR's choices look right

- **Schema versioning instead of name resolution.** A `indexColumns`/schema version in `report.json` would make the layout explicit rather than sniffed. But it needs a matcher change, does nothing for reports already published, and header-name lookup is strictly more tolerant. Current approach is the better one; no change suggested.
- **`getClusterExpansionZoom` for the cluster click** (`report.tsx:222`) would zoom exactly enough to break the cluster. Not worth it here: `clusterMaxZoom: 10` means the `max(zoom + 2, 11)` floor already declusters fully in one click, synchronously.
- **Keeping the report's stops visible under the preview overlay**, instead of filtering them all out. The filter-based hide is the right call for the source-rebuild cost; if two layers were ever wanted at once, the current shape supports it without another teardown path.

### Typos in `selection-info.tsx` (pre-existing, all user-visible)

`Gtfs Feautures` / `OSM Feautures` / `New OSM Feautures` → `Features` (L207, L379, L386, L398), and `This features were …` → `These features were …` (L380, L405).

---

src/uielements/preview.tsx
```diff
@@ -0,0 +1,335 @@
...
+                console.error('Could not load the variants of route', route.id, e);
+                if (!cancelled) setLoadingVariants(false);
+            });
+
+        return () => { cancelled = true; };
+    }, [selectedRouteId, routes, reportRegion]);
+
+    // Memoised because RoutesMap's effect depends on this array's identity: a fresh one on
+    // every render wipes the drawn line and re-uploads it.
+    const fullRouteEntries = useMemo<FullRouteDisplayEntry[]>(() => {
```
biodranik, line 178: This `useMemo` cannot have the effect its comment claims, because `RoutesMap` rebuilds the array again on its own:

```js
// routes.tsx:21-33
const fullRouteFeatures = fullRoutes.map(...);   // fresh array every render
const allFeatures = fullRouteFeatures;
useEffect(..., [map, allFeatures, mapStylesReady]);
```

`RoutesMap` re-renders whenever `Preview` does (clicking a route pill, `setLoadingVariants`, any App-level re-render), `allFeatures` gets a new identity, and the effect's cleanup wipes the `routes` source to an empty `FeatureCollection` before re-`setData`ing it — the flash is still there.

The memo belongs in `routes.tsx`, where it fixes both callers at once:

```js
const allFeatures = useMemo(() => fullRoutes.map(({ routeKey, coordinates }) => ({ ... })), [fullRoutes]);
```

With that in place this memo and `route-list.tsx:133` become load-bearing; without it neither does anything, and the comment above sends the next reader to the wrong file.

---

src/services/matchIndex.ts
```diff
@@ -0,0 +1,168 @@
...
+    for (let i = 1; i < lines.length; i++) {
+        const line = lines[i];
+        if (!line) continue;
+        const c = line.split('\t');
+        // A row truncated before a required column cannot be placed or fetched.
+        if (c.length <= lastNeeded) continue;
+
+        rows.push({
+            id: c[at['gtfs:id']],
+            status: at['status'] !== undefined ? c[at['status']] : '',
```
biodranik: `IndexRow.status` is now dead: `grep -rn "\.status\b\|status:" src/` returns only the type declaration (L91) and this assignment. `category`/`matchedGroup` went in the last round, and `status` was the only other consumer of the column.

Dropping the field, this line, L88-91's doc and the `status` clause in the header comment (L11) removes one optional column from the format contract.

---

src/services/matchIndex.ts
```diff
@@ -0,0 +1,168 @@
...
+            code: c[at[categoryColumn]],
+            type: c[at['type']],
+            lon: parseFloat(c[at['lon']]),
+            lat: parseFloat(c[at['lat']]),
+            byteStart: parseInt(c[at['byte_start']], 10),
+            byteEnd: parseInt(c[at['byte_end']], 10),
```
biodranik: The header is validated loudly, the cells are not — so the failure this function was written to remove survives one row down:

- L137 drops a truncated row with no word to anyone.
- `parseFloat`/`parseInt` are unchecked: a non-numeric `lon` gives a `NaN` coordinate (MapLibre drops the feature), a non-numeric `byte_start` gives `bytes=NaN-NaN` and a 416 on click.

Both land in the same place the PR set out to leave empty: a map with fewer stops than the report has, and nothing said. Counting them into the existing channel is two lines:

```js
let skipped = 0;
...
if (c.length <= lastNeeded || !Number.isFinite(lon) || !Number.isFinite(byteStart)) { skipped++; continue; }
```

and returning/throwing the count so `report.tsx` can show it.

---

src/uielements/report.tsx
```diff
@@ -245,27 +202,68 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
         if (!feature) return;
         const p = feature.properties;
         const [lon, lat] = (feature.geometry as { coordinates: number[] } & any)?.coordinates || [p.lon, p.lat];
+        // The banner describes the attempt in progress, not every attempt since the
+        // region loaded; a stale one reads as if the report itself were broken.
+        setLoadError(null);
```
biodranik: `loadError` now carries two messages with different lifetimes, and the short one erases the long one.

The unknown-category warning (L157) is a property of the loaded index — true until the region changes. This `setLoadError(null)` (and the one at L241) is about a single click. So: open a report this build cannot fully read, click any stop, and the banner explaining why stops are missing is gone for the rest of the session, with the map still missing them.

Two states keep each message on its own clock:

```js
const [indexError, setIndexError] = useState<string | null>(null);   // per region
const [actionError, setActionError] = useState<string | null>(null); // per attempt
```

and render both.

---

src/uielements/report.tsx
```diff
@@ -525,8 +536,8 @@ type DatasetMapLayerProps = {
     data: GeojsonDataT
     onClick?: (datasetName: string, feature?: MapGeoJSONFeature, e?: MapLayerClickEvent) => void
 }
-// Used only for the preview overlay (preview.geojson), which is unchanged by the
-// index/NDJSON migration.
+// The preview overlay: one clustered symbol layer over ir-preview.ndjson. Clustered
+// because it holds every stop of the region at once, matched or not.
 function DatasetMapLayer({ name, color, data, onClick }: DatasetMapLayerProps) {
```
biodranik: Simplification: `DatasetMapLayer` has exactly one call site now (L388), always `name='preview'` and `color=PREVIEW_COLOR`. The two props are generality nothing uses, and they leak outwards — `handlePreviewSelect` takes a `_name` it ignores, and the source/layer/icon ids are built by string concatenation from a constant.

Inlining them as `PreviewLayer({ data, onClick })` with `stops-preview` literals drops both props, the unused callback parameter and the `onClick(name, ...)` signature. The comment on L539-540 already describes it as the preview overlay rather than a generic one.

---

src/uielements/preview.tsx
```diff
@@ -0,0 +1,335 @@
...
+                        {selectedRouteId && (
+                            loadingVariants ? (
+                                <div>Loading route variants&hellip;</div>
+                            ) : variants.length > 0 ? (
+                                <div>
+                                    <RoutesMap fullRoutes={fullRouteEntries} />
+                                    {variants.length > 1 && (
+                                        <div>
+                                            {variants.map((v, i) => (
+                                                <span key={v.inx} className={cls('route-variant')}>
```
biodranik: `.route-variant` carries `cursor: pointer` (`route-list.css:44`) because in `RouteList` the variants are clickable. Here they are labels with no handler, so they advertise a click that does nothing. Either drop the class and style the labels in `preview.css`, or make them select a variant as `RouteList` does.

Also `cls('route-variant')` with one static argument is just `"route-variant"`.

---

src/services/matchIndex.ts
```diff
@@ -0,0 +1,168 @@
...
+export const CATEGORIES: { [code: string]: Category } = {
+    mat: { group: 'matched', label: 'match', color: 'green', help: 'Matched to one or more OSM features; the badges say how' },
+    // Older reports split a match across these four by strategy, one row each; current
+    // ones use `mat` and name the strategies separately. Kept so an old report still renders.
+    mid: { group: 'matched', label: 'match-id', color: 'green', help: 'Stop matched by GTFS id or code' },
+    mrt: { group: 'matched', label: 'match-routes', color: 'green', help: 'Stop matched by routes going through it' },
+    mnm: { group: 'matched', label: 'match-name', color: 'green', help: 'Stop matched by name' },
+    nic: { group: 'matched', label: 'name-id-conflict', color: 'green', help: 'Stop matched by name but mismatched by id' },
+    gen: { group: 'matched', label: 'match-generic', color: 'green', help: 'Matched to a nearby OSM stop that has neither a name nor a code' },
+    sep: { group: 'matched', label: 'separated-cluster', color: '#467d18', help: 'Many OSM stops matched one or many GTFS stops, but successfully separated' },
```
biodranik: Proofreading, now that this is the single copy: `sep`'s "…, but successfully separated" leaves the final clause without a subject — it reads as if the GTFS stops were separated. Suggest **"Many OSM stops matched one or many GTFS stops, and were successfully separated"**.

The rest of the table and `STRATEGIES` read well; `platform`'s `local_ref`/`ref` phrasing is the right OSM vocabulary.

---

src/uielements/selection-info.tsx
biodranik, line 1: Three items in this file that have no line inside a diff hunk:

**1. L514 is the third writer of a selection hash and it was not encoded.** `app.tsx:139` and `preview.tsx:207` now percent-encode the id; the "already matched to another gtfs feature" links still interpolate `m` raw:

```js
<a href={`/#/match-report/${reportRegion}/selection/${m}`}>
```

`m` is a GTFS id from `f.mtch`, so a `#` truncates the link and a `/` is cut by `parseSelectionHash`'s `[^/]+` — exactly the bug the PR fixes elsewhere. `encodeURIComponent(m)` closes it.

**2. `DatasetHelp` (L569) now renders the same string the category tag already carries.** L180 puts `matchInfo.help` in the tag's `title`, L189 prints `CATEGORIES[datasetName].help` as visible italics — same text, twice on screen. Keeping the paragraph and dropping the `title` (or the reverse) removes the component or the attribute; either way one of them is redundant.

**3. Capitalisation.** With the labels being consolidated, `Gtfs stop Id` / `Gtfs stop Code` (L194-195) and the `Gtfs Features` heading (L199) are the odd ones out — everything else in the new tables says `GTFS`, and `Id` should be `ID`.

---

biodranik (commented):
Re-reviewed on `ab90ac1`. Every thread from the previous round is addressed — `memoFetch`, the `hashchange` cleanup, `preview.css`, the cancellation guard, the unknown-category banner, the `alias` collapse and all the wording fixes landed as suggested; resolving those. Eight new items inline, plus one that spans files:

**The route-line re-upload is not fixed by the memo added here.** `routes.tsx` rebuilds `allFeatures` on every render of its own, so `preview.tsx:178` and `route-list.tsx:133` are both currently no-ops. Details on the `preview.tsx` thread — the one-line fix is in `routes.tsx`.

### Alternatives considered, and why the PR's choices look right

- **Bit field vs. name list for `index.tsv`'s `strategies`.** Now that nothing decodes the column, the eagerly-loaded index pays nothing for it either way and the choice stays open for whoever builds the facet. Leaving it undecoded is the right call for this PR; no change suggested.
- **Deriving badges from the index instead of the detail body.** The body names the strategies and is fetched only for the selected stop, so the badges cost one range request that was already happening. Cheaper than widening the per-region file. Current approach is better.
- **`strategiesOf` recomputed per render, `STRATEGIES.find` per badge.** Seven entries, one selected stop — memoising would cost more than it saves. Correct as written.
- **Header-name resolution vs. a schema version in `report.json`.** Unchanged from the last round: name lookup is strictly more tolerant and needs no matcher change. Agreed.

---

src/uielements/report.tsx
kiselev-dv: `IndexRow.status` was the last of the four and it is gone now, along with the `status` clause in the header comment. `grep -rn "matchedGroup\|[^A-Za-z]category" src/` filtered for `CATEGOR`/`subcategory` returns nothing.

src/uielements/preview.tsx
kiselev-dv: You are right — it was a no-op, and the comment sent the reader to the wrong file. Moved into `routes.tsx`, memoised on `[fullRoutes]`, which fixes both callers at once and makes this memo and `route-list.tsx:133` load-bearing rather than decorative. The comment there now says why it lives in that component: it re-renders whenever its parent does, so the call sites cannot stabilise the identity from outside.

src/services/matchIndex.ts
kiselev-dv: Dropped — the field, the assignment, its doc comment and the `status` clause in the header comment. One fewer optional column in the format contract.

src/services/matchIndex.ts
kiselev-dv: Counted, close to your sketch. `parseIndex` returns `{ rows, skipped }` and rejects a row whose `lon`, `lat`, `byte_start` or `byte_end` is not finite, alongside the truncation case. `report.tsx` folds the count into the index banner beside the unknown-category message, so `index.tsv: 3 rows could not be read` shows up on load rather than as three stops nobody notices are missing.

The javadoc spells out the two failure shapes, since neither is obvious from the code: a NaN coordinate MapLibre silently drops, and a `bytes=NaN-NaN` request that 416s on click.

src/uielements/report.tsx
kiselev-dv: Split into `indexError` and `actionError`, both rendered. The index one is set on load (unknown categories, unreadable rows, a failed fetch) and cleared only when the region changes; the action one is cleared at the start of each click and set on its failure. The deep-link path at L299 was a third writer of the same state and now uses the action channel.

src/uielements/report.tsx
kiselev-dv: Inlined as `PreviewLayer({ data, onClick })`. `stops-preview` and `stop-preview` are literals, `PREVIEW_COLOR` is read directly, `handlePreviewSelect` loses the `_name` it ignored, and the `onClick` signature is `(feature, e)`.

src/uielements/preview.tsx
kiselev-dv: Took the first option — these are labels, so they should not look clickable. New `.preview-route-variant` in `preview.css`, with a comment saying why it is not `.route-variant`. `cls()` with one static argument is gone.

src/services/matchIndex.ts
kiselev-dv: Taken verbatim: "Many OSM stops matched one or many GTFS stops, and were successfully separated".

src/uielements/selection-info.tsx
kiselev-dv: All three done.

1. `encodeURIComponent(m)` on the `f.mtch` links, with a comment naming the other two writers so the three stay in step.
2. Dropped the `title` from the category tag and kept `DatasetHelp`'s visible paragraph — the visible one is the more useful of the pair. Left a note on the `<h2>` so nobody adds the attribute back.
3. `GTFS stop ID`, `GTFS stop code`, `GTFS Features`.

kiselev-dv:
Addressed in `6a154d1`; all threads replied to and resolved.

One fix beyond the review, for a doubling of the route lines on the map. `LayerControls.addOverlayImmediate` pushed unconditionally, while `RoutesMap` guarded on `map.getSource('routes')` — two different registries. `routesStyling` is a module-level constant, so once it is registered twice, `setBaseStyle` replays its three layers over the one source and the line, its arrows and its labels are all drawn on top of themselves. `map.getSource` reads absent for a moment while `setStyle` rebuilds, which is the window.

`addOverlayImmediate` is idempotent now (identity check on the overlay, plus `getSource`/`getLayer` guards), `removeOverlayImmediate` is defensive to match, and `RoutesMap` no longer guards on the map.

Flagging it because I have not reproduced it in a browser — the mechanism is clear from the code and the fix is right either way, but if what you saw was a route drawn twice **while the base style has not been cycled**, that is a different cause and worth another look. The other candidate is `route-list.tsx` drawing every variant of every route when no route is selected, so a bus route's two directions overlap along the same road.

PR #158, opened 2026-08-29
Add agent instructions and refresh docs

biodranik:
## What changed

**Commit 1 — agent instructions**
- `CLAUDE.md`: guidance for AI coding agents — data contract (index.tsv + NDJSON Range fetches, routes, schedule format v4), architecture, conventions, error-handling policy. `AGENTS.md` is a symlink to it (same convention as the main Organic Maps repo).

**Commit 2 — docs refresh to match the current code**
- README + in-app Help overlay: added the missing match categories (`match-routes`, `name-id-conflict`, `match-generic`), renamed categories to the current UI labels (`separated-cluster`, `transit-hub`, `cluster`, `no-osm`), documented the built-in OSM editor workflow (edit/move/add stop → download `.osm` from the OSM Changes tab), fixed typos.
- `docs/component-tree.md`: added `RouteList` and `MatchArrowLayer`, corrected editor component placement, replaced the retired per-dataset layer model with `StopsLayer`, refreshed the context consumers table.

## Testing

`yarn build` (tsc + vite) passes. Doc content was verified against the sources; category names and descriptions come from `CATEGORIES`/`MATCH_LABELS` in the code.

Written with an LLM agent, reviewed by me.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,90 @@
+This file provides guidance to AI agents when working with code in this repository. `AGENTS.md` is a symlink to this file (same convention as the main Organic Maps repo). See also [README.md](README.md) (project purpose, match categories, contribution info) and [docs/component-tree.md](docs/component-tree.md) (full component/layout tree — keep it updated when moving components).
+
+## What this is
+
+A browser SPA for the Organic Maps project that visualizes and validates the matching of **GTFS public-transport stops to OpenStreetMap stops**, and helps fix mismatches: besides inspecting pre-computed match results on a map, a user can edit OSM stop tags, move stops, create new ones, and download the edits as a JOSM-compatible `.osm` file. It also previews GTFS timetables with live (GTFS-RT) delays.
+
+Data sources (no backend of its own):
+- Static match data under `DATA_BASE_URL` (`src/config.ts`: `VITE_DATA_BASE_URL` env, prod `https://pt.organicmaps.app/data` via `.env.production`, dev default `/data` proxied by Vite) — produced by the sibling `../gtfs-parser` project (separate repo).
+- Schedules API: `http://localhost:4567/v1/{schedule,updates}` in dev, `https://pt.organicmaps.app/api/v1/…` in prod (hardcoded in `schedule-preview.tsx`).
+- Live OSM data: Overpass API (per-tile stop queries) and `api.openstreetmap.org/api/0.6` (elements by id) via `src/services/OsmQuerryQueue.ts`.
+- Map tiles: OpenFreeMap "bright" style + Esri world imagery (`src/map/styling.ts`; the satellite style borrows OpenFreeMap sprite/glyphs so symbol layers keep working).
+
+Stack: Preact 10 + Vite 7 + TypeScript 5.9 (strict) + MapLibre GL 5. **Yarn 4 via corepack** (`packageManager` in package.json).
+
+## Commands
+
+```bash
+yarn dev          # Vite dev server (proxies /data -> http://localhost:8801, stripping the /data prefix)
+yarn serve-data   # http-server for ../gtfs-parser/data/match on :8801 --cors
+yarn build        # tsc -b && vite build  (type-check gates the build; also what CI runs)
+yarn preview      # serve the production build
+```
+
+CI (`.github/workflows/build.yml`) runs `yarn install --immutable && yarn build` on every push/PR. `yarn deploy` runs `./scripts/deploy.sh` — `scripts/` is deliberately gitignored (local-only), so don't rely on it. Deployment is Cloudflare static assets from `./dist` (`wrangler.jsonc`); a custom Vite plugin (`selectiveOutDirCleanup` in `vite.config.ts`) preserves `dist/data` across builds so match data can be colocated.
+
+There are no tests and no ESLint. The only static check is `tsc -b`. The `tsconfig`s are strict and include `noUnusedLocals`/`noUnusedParameters`/`noFallthroughCasesInSwitch`, so dead code and unused params **fail the build** — clean these up rather than leaving them.
+
+## Data contract (the heart of the app)
+
+- `GET {DATA_BASE_URL}/match-report.json?t=<now>` → `{ matchedRegions: Report[], foldByName: string[] }` — region list with per-region `matchStats`, `idTags` (which OSM tags hold GTFS ids, with counts — drives the editor's Set Id/Code defaults), `matchMeta` (timestamps, bbox, sources), `liveUpdates`; `foldByName` holds region-name prefixes the report table folds into groups. See `Report` in `src/uielements/report.tsx`. (There is no separate update-report.json anymore.)
+- `GET {DATA_BASE_URL}/{region}/index.tsv?t=<now>` → one row per GTFS stop: `id, status (m|…), code, type, lon, lat, byteStart, byteEnd, search_terms(ignored)`. `code` is the 3-letter sub-category (`status_detailed`), `type` picks the detail file.
+- Stop details are **NDJSON fetched with HTTP `Range` headers**: `type` → `matches.ndjson` / `clusters.ndjson` / `no-osm.ndjson` (`fileFor` in `report.tsx`), byte offsets from the index row. The detail JSON becomes the selected feature's properties.
+- `GET {DATA_BASE_URL}/{region}/routes.ndjson` → route index (with `byteOffset`/`byteLength`); `route-stops.ndjson` is then Range-fetched per route for variants + polylines (`route-list.tsx`, with module-level promise caches).
+- `GET {DATA_BASE_URL}/{region}/preview.geojson` → stops for the timetable-preview overlay.
+- Schedules API: `GET …/schedule/{stopId}` returns `formatVersion: '4'` — a compact stop-centric format (deduplicated arrays, VLQ/base64url `pos` blobs, bit-vector service periods) decoded by `src/services/ScheduleEncoding.ts`; the format is documented in `src/services/schedule.types.d.ts`. `GET …/updates/{stopId}` (GTFS-RT delays) is polled every 5 s only when `feedMeta.liveUpdates`, and gives up after 3 errors.
+
+**Match categories** are keyed by the 3-letter `code`: `CATEGORIES` in `report.tsx` (group matched/not-matched, label, color, help) and `MATCH_LABELS` in `selection-info.tsx`. Adding/renaming a category means touching both, and the pipeline that writes index.tsv.
+
+GeoJSON feature `properties` carry nested data **as JSON strings** (e.g. `osmFeatures`, `gtfsFeatures`, `gtfsRoutes`). `stringifyProperties` in `report.tsx` enforces this when a feature flows into selection state, and the panels `JSON.parse` it back (`parseJsonSafe`). Preserve this string-encoding when adding fields — MapLibre feature properties can't hold nested objects.
+
+Gotcha: in `matchMeta.gtfsBbox`, `top` is the **south** latitude and `bottom` the north (see `region-markers.tsx`).
+
+## Architecture
+
+Component/layout tree, side-panel tabs, and context consumers are documented in [docs/component-tree.md](docs/component-tree.md). The essentials:
+
+**Global state** (`src/app.tsx`):
+- `MapContext` → `{ map, loaded: Promise<Map>, layerControls }`. Await `loaded` before adding images/overlays — don't race `map.loaded()`.
+- `SelectionContext` → `{ selection, selectionSource, updateSelection(sel, source), onReportSelect }`. `selectionSource` (`'map-click' | 'url-hash' | …`) matters: e.g. only `url-hash` selections trigger a flyTo.
+- `OSM_DATA` (`src/services/OSMData.ts`) — singleton store of live OSM elements + tracked edits (create node with negative id, move, change tags), consumed via `subscribe` + `useSyncExternalStore`. `OSM_QUERY_QUEUE` fills it from Overpass (z16 tiles, deduped, 1 s throttle) and the OSM API.
+
+**Hash routing, bidirectionally synced** (`src/uielements/routing.ts`):
+- URL shape: `#/match-report/{region}` plus `/selection/{id}` or `/preview/{id}`. The category is **not** in the URL — deep links recover it from the region's index.tsv row.
+- `useHash()`/`useHashRoute(parser)` are the read direction; a `useEffect` on `selection` in `app.tsx` writes the hash. Keep both directions in mind when touching selection logic.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Purpose of this project
 
-In order to display GTFS stops on the map without doubling ond visual clutter
+In order to display GTFS stops on the map without doubling and visual clutter
 we need to match GTFS stops to OSM stops.
 
 This UI displays matching results in more convenient way.
```
biodranik: `without doubling and visual clutter` still doesn't read — the intended sense is duplicates:

```suggestion
In order to display GTFS stops on the map without duplicates and visual clutter
we need to match GTFS stops to OSM stops.

This UI displays matching results in a more convenient way.
```

Same paragraph, l.11: `with different level of confidence` → `with different levels of confidence`.

---

README.md
```diff
@@ -55,45 +55,60 @@ Please take a look what tag is considered appropriate by the local community.
...
+### match-routes
+These stops were matched by the routes going through them.
+
 ### match-name
-This stops were matched by name and type and didn't get into a cluster of matches.
-Names are getting normalised: Special characters removed, lowercase, diacritics removed,
-ß converted to ss etc. Names are checked against *name* elemnt tags.
+These stops were matched by name and type and didn't get into a cluster of matches.
+Names are getting normalised: special characters removed, lowercase, diacritics removed,
+ß converted to ss etc. Names are checked against *name* element tags.
```
biodranik: The normalisation sentence mixes a passive list with an adjective (`special characters removed, lowercase, diacritics removed`):

```suggestion
Names are normalised: special characters and diacritics are removed, names are lowercased,
ß is converted to ss, etc. Names are checked against *name* element tags.
```

Same text in `report-help-overlay.tsx` l.32-33.

---

README.md
```diff
@@ -55,45 +55,60 @@ Please take a look what tag is considered appropriate by the local community.
...
 and any number of stops or platforms.
 
 ### many-to-one
-This are clusters where multiple GTFS stops were matched to excatly one OSM element.
+These are clusters where multiple GTFS stops were matched to exactly one OSM element.
 
-### clusters
-This are clusters which the tool was unable to separate.
+### cluster
+These are clusters which the tool was unable to separate.
```
biodranik, line 105: Two things about the cluster categories:

1. **`many-to-one` is described in opposite directions in the repo.** Here (and in the help overlay): "multiple GTFS stops were matched to exactly one OSM element". The tooltips users see say the reverse — `report.tsx:36` `mto: 'Many OSM stops matched exactly one GTFS by name'` and `selection-info.tsx:69` `'Many OSM stops matched exactly one GTFS stop by name'`. `transit-hub` in both places is OSM-side-many (one hub + N platforms), which makes the tooltip direction the more likely correct one, but `DatasetHelp` (`selection-info.tsx:570`) says GTFS-side-many for the whole `clu/mto/hub/sep` group. Worth settling against the pipeline and making all three agree — a docs PR is the right place for it.

2. **Order.** The overlay was reordered to the `CATEGORIES` order (`sep, clu, mto, hub`), README kept `sep, hub, mto, clu`. Same order in both would make them readable side by side.

---

src/uielements/report-help-overlay.tsx
```diff
@@ -6,56 +6,65 @@ export function ReportHelpOverlay({ onClose }: { onClose: () => void }) {
...
+                    Main goal of this tool is to link GTFS stops to OSM stops.
 
-                    Otherwise map would be cluttered with stps from different sources.
+                    Otherwise the map would be cluttered with stops from different sources.
                 </p>
                 <p>
                     There are different ways how to establish that match with different level of confidence.
-                    Different sets shows different groups of matches or errors if tool can't match
+                    Different sub-categories show different groups of matches or errors if the tool can't match
                     GTFS and OSM stop with enough confidence.
```
biodranik, line 16: Two grammar leftovers in the intro:

```suggestion
                    The main goal of this tool is to link GTFS stops to OSM stops.

                    Otherwise the map would be cluttered with stops from different sources.
                </p>
                <p>
                    There are different ways to establish that match, with different levels of confidence.
                    Different sub-categories show different groups of matches or errors if the tool can't match
                    GTFS and OSM stop with enough confidence.
```

---

src/uielements/report-help-overlay.tsx
```diff
@@ -6,56 +6,65 @@ export function ReportHelpOverlay({ onClose }: { onClose: () => void }) {
...
                 </p>
                 <p>
                     <ul>
```
biodranik, line 19: Same class of errors is still visible elsewhere in the UI — cheap to fix in this pass:

- `report.tsx:34` — `sep` help: `successfuly` → `successfully` (`selection-info.tsx:67` already has it right).
- `report.tsx:38` — `nom` help: `No osm element matched` → `No OSM element matched` (matches `MATCH_LABELS`).
- `selection-info.tsx:345` — `This features were just created` → `These features were just created`.
- `selection-info.tsx:370` — `This features were not considered as match candidates during server matching` → `These features …`.

---

CLAUDE.md
```diff
@@ -0,0 +1,90 @@
...
+- Schedules API: `GET …/schedule/{stopId}` returns `formatVersion: '4'` — a compact stop-centric format (deduplicated arrays, VLQ/base64url `pos` blobs, bit-vector service periods) decoded by `src/services/ScheduleEncoding.ts`; the format is documented in `src/services/schedule.types.d.ts`. `GET …/updates/{stopId}` (GTFS-RT delays) is polled every 5 s only when `feedMeta.liveUpdates`, and gives up after 3 errors.
+
+**Match categories** are keyed by the 3-letter `code`: `CATEGORIES` in `report.tsx` (group matched/not-matched, label, color, help) and `MATCH_LABELS` in `selection-info.tsx`. Adding/renaming a category means touching both, and the pipeline that writes index.tsv.
+
+GeoJSON feature `properties` carry nested data **as JSON strings** (e.g. `osmFeatures`, `gtfsFeatures`, `gtfsRoutes`). `stringifyProperties` in `report.tsx` enforces this when a feature flows into selection state, and the panels `JSON.parse` it back (`parseJsonSafe`). Preserve this string-encoding when adding fields — MapLibre feature properties can't hold nested objects.
+
+Gotcha: in `matchMeta.gtfsBbox`, `top` is the **south** latitude and `bottom` the north (see `region-markers.tsx`).
```
biodranik, line 41: This gotcha canonises a claim the code contradicts, so it will mislead agents either way:

- `region-markers.tsx:36` asserts it, but nothing there depends on it — a rectangle's corners and the centre `(top + bottom) / 2` come out the same under either orientation, so that comment is unverified by its own code.
- `report.tsx:186-190` is the only orientation-sensitive use and assumes the **opposite**: `fitBounds([[left, bottom], [right, top]])` = `[SW, NE]`, i.e. `bottom` = south.

If the gotcha holds, "select a region" fits an inverted box (maplibre takes the pair as sw/ne without normalising, so `north < south` gives a negative box height). Suggest checking one `matchMeta.gtfsBbox` in `match-report.json`, then fixing whichever side is wrong and documenting the verified one — or dropping both the comment and this line if the field is plain `top` = north.

---

CLAUDE.md
```diff
@@ -0,0 +1,90 @@
...
+
+There are no tests and no ESLint. The only static check is `tsc -b`. The `tsconfig`s are strict and include `noUnusedLocals`/`noUnusedParameters`/`noFallthroughCasesInSwitch`, so dead code and unused params **fail the build** — clean these up rather than leaving them.
+
+## Data contract (the heart of the app)
+
+- `GET {DATA_BASE_URL}/match-report.json?t=<now>` → `{ matchedRegions: Report[], foldByName: string[] }` — region list with per-region `matchStats`, `idTags` (which OSM tags hold GTFS ids, with counts — drives the editor's Set Id/Code defaults), `matchMeta` (timestamps, bbox, sources), `liveUpdates`; `foldByName` holds region-name prefixes the report table folds into groups. See `Report` in `src/uielements/report.tsx`. (There is no separate update-report.json anymore.)
```
biodranik, line 30: `(There is no separate update-report.json anymore.)` describes what the contract *used to* be — by this file's own rule (l.85: never reference the previous implementation) it should just state the current contract. Drop the sentence.

Same pattern in code comments, if you want them consistent with the rule this PR introduces: `report.tsx:528-529` ("unchanged by the index/NDJSON migration"), `report.tsx:272` and `routing.ts:44` ("no longer encoded in the URL"). The useful half of the last two — "the category is recovered from index.tsv" — is worth keeping.

---

CLAUDE.md
```diff
@@ -0,0 +1,90 @@
...
+- Types use a `…T` suffix (`SelectionT`, `FeedMetaT`); newer service types drop it (`Schedule`, `OSMElement`) — match the file's neighbors. `cls()` (`uielements/cls.ts`) joins conditional class names.
+- No formatter is configured; indentation varies by file (2 spaces in `app.tsx`, 4 in most others) — match the file you're editing.
+- Comments should be brief, explaining only reasoning that is not obvious from the code itself.
+- A few `@ts-ignore`s exist around MapLibre's style typings; that's tolerated there, not an invitation to add more elsewhere.
+
+## Error handling policy
+
+Match data comes from our own pipeline and is trusted: fail visibly in the console rather than adding defensive fallbacks or schema validation for inputs that can only be wrong due to a pipeline bug (`parseJsonSafe` exists only for genuinely optional properties). Third-party live services (Overpass, OSM API, schedules API) **do** get guarded: catch, log, and degrade (see the trip-updates 3-strike cutoff and `OsmQuerryQueue`'s try/catch + throttle) — never hammer them in a retry loop.
```
biodranik, line 78: "`OsmQuerryQueue`'s try/catch" only holds for `queryStopsForTiles`. `queryDataByIds` (`OsmQuerryQueue.ts:88-127`) is bare `fetch().then()`, and its only caller doesn't guard either: `selection-info.tsx:270-281` does `setLoading(true)` → `await queryDataByIds(...)` → `setLoading(false)` with no try/finally, so one OSM API failure (`.json()` on an error page throws) leaves "Loading OSM data..." on screen forever plus an unhandled rejection.

The policy in this paragraph is the right one — the code just doesn't meet it yet. Smallest fix is a `try/catch` + `console.error` in `queryDataByIds` mirroring `queryStopsForTiles`, which also makes `setLoading(false)` unreachable-by-throw.

---

CLAUDE.md
```diff
@@ -0,0 +1,90 @@
...
+- Render-less overlay components follow one idiom: build the overlay spec, `mapLoaded.then(...)` guarded by a `subscription = { canceled, promiseFulfiled }` object, clean up in the effect teardown. See `StopsLayer`, `DatasetMapLayer` (preview), `RegionMarkersLayer`, `MatchArrowLayer`, `RoutesMap` (which instead keeps a persistent `routes` source and `setData`s into it), `HtmlMapMarker` (`editor/map-marker.tsx`).
+
+**Selection panel** (`selection-info.tsx`): unified `MatchInfo` for all categories (clusters are just `gtfsFeatures.length > 1`), with `RouteList` (route pills + variants drawn via `RoutesMap`), match arrows, distances (`src/map/distance.ts`), and the editor: `TagEditor` (`editor/osm-tags.tsx`), Set Name/Id/Code quick actions, `MoveController`, `AddOsmStopController` (creates a node with proper PT tags per route type), plus "Surrounding OSM Features" from Overpass within 500 m. Edits accumulate in `OSM_DATA.changes`; the Changes tab (`editor/changes.tsx`) exports them as `gtfs-changes.osm` (JOSM-loadable XML). An `mtch` field on an OSM candidate warns it is already matched to another GTFS stop.
```
biodranik, line 63: Documenting `idTags` as what "drives the editor's Set Id/Code defaults" surfaces two problems in that code:

1. `add-stop-controller.tsx:34` picks `Object.entries(idTags)[0]` **without** the `k !== 'name'` filter that `selection-info.tsx:232` applies. When `name` is the first key, the new node gets `{ name, ['name']: code }` — the stop name is silently overwritten by the GTFS code and no ref tag is written at all. Reuse the same filtered lookup in both places.
2. `selection-info.tsx:233-235`: Set Id and Set Code write the **same** `gtfsIdTag` key, so Set Code overwrites the id tag. If that's intended (regions using one ref tag for both), a one-line comment would help; otherwise Code should prefer an id tag matching `/code/`.

Both also take `[0]` of an object whose values are usage counts — picking the highest count would be a more meaningful default, and is the same amount of code.

---

docs/component-tree.md
```diff
@@ -66,8 +70,8 @@ App  (src/app.tsx)
 
 | Context | Provider | Consumers |
 |---|---|---|
-| `MapContext` | `App` | `MatchReport`, `DatasetMapLayer`, `RegionMarkersLayer`, `RoutesMap`, `SelectionInfo` |
-| `SelectionContext` | `App` | `MatchReportSelector`, `MatchReport` |
+| `MapContext` | `App` | `MatchReport`, `StopsLayer`, `DatasetMapLayer`, `RegionMarkersLayer`, `RoutesMap`, `MatchArrowLayer`, `HtmlMapMarker`, `LocateMe`, `AddOsmStopController`, `MoveController` |
```
biodranik, line 73: `LocateMe` is now cited here but doesn't appear in the tree above — add it where it renders (`SchedulePreview` header, `schedule-preview.tsx:230`, and `OsmListElement`, `selection-info.tsx:505`) so the table and the tree agree.

---

biodranik (commented):
Docs check out against the code: the data contract (index.tsv columns, `fileFor`, Range-fetched NDJSON, routes/route-stops, schedule v4 + 5 s poll + 3-strike cutoff), the category list (all 11 codes, current labels), the context-consumer table (matches `useContext` usage exactly), and the component tree all match. Findings are inline; two are worth calling out:

- The `gtfsBbox` gotcha (CLAUDE.md l.41) contradicts `report.tsx`'s `fitBounds` — one of the two is a bug.
- `many-to-one` is documented in opposite directions in README/overlay vs. the in-app tooltips.

**Simplification worth doing (separately or here).** The category descriptions live in five places — README, the help overlay, `CATEGORIES.help` (`report.tsx:28-40`), `MATCH_LABELS` (`selection-info.tsx:61-73`) and `DatasetHelp`'s if-chain (`selection-info.tsx:541-574`) — which is exactly why they have drifted. `MATCH_LABELS` is a near-copy of `CATEGORIES`: same 11 codes, same labels, only the help wording differs. Exporting `CATEGORIES` and deleting `MATCH_LABELS` removes one full list and makes CLAUDE.md l.37's "adding a category means touching both" untrue by construction. Going further — a second `description` field on `CATEGORIES` and rendering the overlay's `<li>`s from it — removes a third copy and shrinks `report-help-overlay.tsx` to a loop; the trade-off is one table holding both a short tooltip and long prose. The README copy would stay manual either way, so it's the one place worth keeping a link to the in-app help instead of duplicating the text.

---

Issue #140, opened 2026-06-26
Some french GTFS


rmezeretteENSG:
Regional / High speed trains
https://eu.ftp.opendatasoft.com/sncf/plandata/Export_OpenData_SNCF_GTFS_NewTripId.zip (from https://transport.data.gouv.fr/datasets/horaires-sncf)

Paris region (larger area than the one already set)
https://eu.ftp.opendatasoft.com/sncf/gtfs/transilien-gtfs.zip (from https://transport.data.gouv.fr/datasets/horaires-des-lignes-transilien-1)

Some Nancy public transportation :
https://transport.data.gouv.fr/resources/80431/download (from https://transport.data.gouv.fr/resources/80431)
https://transport.data.gouv.fr/resources/83710/download (static)
https://proxy.transport.data.gouv.fr/resource/fluo-stan-nancy-gtfs-rt-trip-update (real time)
-> from https://transport.data.gouv.fr/datasets/fr-200052264-t0034-0000-1

For Orleans:
https://chouette.enroute.mobi/api/v1/datas/keolis_orleans/gtfs.zip (static)
https://ara-api.enroute.mobi/tao/gtfs/trip-updates (real time)
-> from https://transport.data.gouv.fr/datasets/gtfs-netex-et-gtfs-rt-reseau-tao-orleans-metropole

Lorient
https://s3.eu-west-1.amazonaws.com/files.orchestra.ratpdev.com/networks/rdla-lorient/exports/medias.zip (static)
https://proxy.transport.data.gouv.fr/resource/ctrl-lorient-gtfs-rt-alerts (real time)
-> from https://transport.data.gouv.fr/datasets/horaires-theoriques-du-reseau-izilo-format-gtfs-de-lorient-agglomeration

Bretagne region
https://www.korrigo.bzh/ftp/OPENDATA/KORRIGOBRET.gtfs.zip (static)
-> from https://transport.data.gouv.fr/datasets/base-de-donnees-multimodale-transports-publics-en-bretagne-korrigo-gtfs

Thanks for your work

PR #163, opened 2026-09-10
Show dissolution and transit POIs on the map

kiselev-dv:
The report has been saying, per stop, that a stop stands in for several places and what
dissolution did about it. Nothing here read that. This shows it, and puts the OSM transit
features the matcher works from on the map beside it.

Server side: organicmaps/gtfs-server#140 and #143, both merged.

## Dissolution

A dissolved stop is **not** a deleted stop, and the panel has to say so or the map teaches the
opposite. The stop keeps its row, its id and its coordinates; what moved is its departures,
dealt to the places it stands for, and any call no part could take stays on it.

- **On the map**, a stop whose calls this run dealt is drawn grey rather than in its match
  category's colour. It stays in that category, stays behind that category's checkbox, and
  stays clickable — the colour says "its departures are elsewhere now", not "it is gone".
- **In the panel**, the parts it was dealt to, split by the two shapes the report distinguishes:
  the stops the feed already named as its parts, and the ids that exist only in the published
  data. Those are different facts about the feed and the panel does not merge them.
- **Both tenses.** `dissolving` is a run that dealt the visits; `dissolvable` is one that
  measured and acted on nothing, which is every feed the operator has not switched on. The word
  comes from the record; which tense a feed is in comes from `matchedRegions[].dissolution`
  (`on` / `analysis` / `off`), stated once per feed rather than repeated on every row.
- **`notDissolved`** names the rule that refused, in the words of someone fixing OSM rather
  than the pipeline's own — "the places it matched are too close together to tell apart", not
  `partsCloserThanMinSeparation`. An unrecognised code shows as itself instead of as nothing.

`index.tsv`'s flag column carries the two bits the map filters on (`1 << 7` planned,
`1 << 9` declined); the panel's words come from the stop's own record, which is fetched only
for the stop someone clicked.

## Transit POIs

`pt-structure.pmtiles`, the stop structure layer the matcher builds, drawn as an optional
overlay: the bays and stop features OSM holds near a feed's stops. It is what makes a bad match
legible — a stop drawn far from every OSM feature that could serve it is a different problem
from one drawn on top of three. Off by default, one dependency (`pmtiles`), and the layer 404s
harmlessly where a region has none.

## Notes

- `tsc -b && vite build` clean; rebased on `main` after the dependabot bumps, `yarn install
  --immutable` passes.
- Two branches follow this one and want reviewing in order: `feat/preview-anchors` (draw a stop
  where the matcher anchored it) and `feat/osm-index` (what the matcher looked at and did not
  use).


CLAUDE.md
```diff
@@ -27,8 +27,9 @@ There are no tests and no ESLint. The only static check is `tsc -b`. The `tsconf
...
+- `GET {DATA_BASE_URL}/{region}/index.tsv?t=<now>` → one row per GTFS stop: `id, status (m|…), code, type, lon, lat, byteStart, byteEnd, search_terms(ignored)`. `code` is the 3-letter sub-category (`status_detailed`), `type` picks the detail file. The `strategies` column is a bit field; only its dissolution bits are decoded here (`1 << 7` planned, `1 << 9` declined — `1 << 8` is retired and left unused), because the map filters on them and cannot fetch a body per stop.
+- `GET {DATA_BASE_URL}/{region}/pt-structure.pmtiles` → the stop-structure overlay, fetched only when the PT structure button is pressed, which is also when `pmtiles` itself is imported. A region without one 404s and the button simply draws nothing.
```
biodranik: The documented path has a `{region}` segment, the code has none:

```ts
url: `pmtiles://${DATA_BASE_URL}/pt-structure.pmtiles`   // src/map/map.ts:167
```

One global archive is what the comment at `map.ts:137` describes ("one archive, served from the same data root"), so it's this line that's wrong — and the sentence that follows it ("A region without one 404s") only makes sense for the per-region shape. Worth fixing here and saying instead that the archive is region-independent, i.e. the layer stays drawn across region switches and covers whatever the pipeline put in it.

---

src/services/matchIndex.ts
```diff
@@ -94,8 +94,35 @@ export type IndexRow = {
...
+ * Which tense they belong to is not here, because it is not a fact about the stop: the run
+ * either acted on this feed or only measured it, and `matchedRegions[].dissolution` in the
+ * report says which, once.
+ *
+ * `1 << 8` is skipped. The report used to set it where a named stop also had an assignment,
+ * back when its measuring run had a detector that could name a family it could not deal;
+ * both runs now plan the same way, so it marked exactly the rows `1 << 7` marks.
+ */
+const FLAG_DISSOLUTION_PLANNED = 1 << 7;
+const FLAG_DISSOLUTION_DECLINED = 1 << 9;
```
biodranik, line 124: `FLAG_DISSOLUTION_DECLINED` / `IndexRow.dissolutionDeclined` are computed for every row of every region and never read — the panel's "Left whole" comes from the detail body's `notDissolved`, and `report.tsx:111` uses only `dissolutionPlanned`. `noUnusedLocals` doesn't catch an object property, so nothing flagged it.

Dropping the constant, the field and its doc line also collapses most of this block: the "the first seven are match strategies; these two…" and the `1 << 8` paragraph exist to explain a two-bit set. One bit needs one line. (CLAUDE.md:31 then loses `1 << 9 declined` too — and its "because the map filters on them" is already untrue of that bit.)

---

src/map/map.ts
```diff
@@ -114,3 +126,174 @@ export async function loadSvgWithColors(svgSrc: string, colors: SelectorToColorP
...
+    'railway',        // subclass: station, halt, tram_stop, subway_entrance, funicular
+    'bus',            // subclass: bus_stop, bus_station
+    'ferry_terminal',
+    'harbor',
+    'aerialway',
+    'airport',
+];
+
+/** The rank layers' own filters, which the class test is added to rather than replacing. */
+const POI_RANK_FILTERS: { [id: string]: unknown } = {
```
biodranik, line 289: This hand-copies OpenFreeMap's rank cuts. The style already ships them and `getStyle()` hands them back, so the table can go entirely:

```ts
for (const layer of style.layers) {
    if ((layer as { 'source-layer'?: string })['source-layer'] !== 'poi') continue;
    if (layer.id === 'poi_transit') continue;   // already transit-only
    const own = (layer as { filter?: unknown }).filter;
    // @ts-ignore — the style's own filter, and the class test beside it.
    map.setFilter(layer.id, own ? ['all', own, transitOnly] : transitOnly);
}
```

Three wins: ~12 lines gone; upstream renaming a layer or moving a rank boundary can't silently replace its filter with our stale copy; and a *new* `poi` layer gets decluttered instead of leaking clutter back (today `if (own)` skips anything unknown).

Two more while here:
- `hidePoiLayers` no longer hides anything — `keepTransitPoisOnly` says what it does.
- Alternative worth considering: `styling.ts` already exports `mixinToStyle` (currently unused) and `LayerControls.resolveStyles()` is where the URL becomes a `StyleSpecification`. Filtering there bakes it into the style, so it survives switches by construction and the `style.load` handler and its four-line justification disappear.
- Product question: this is now unconditional with no way back. A mapper checking a shelter, a P+R or a landmark shop against the base map has lost it. Hanging it off a toggle (or the Sat/Geo cycle) would keep both.

---

src/map/map.ts
```diff
@@ -114,3 +126,174 @@ export async function loadSvgWithColors(svgSrc: string, colors: SelectorToColorP
...
+        // Marked before the awaits: the button answers the finger, and the layer catches up.
+        button.classList.toggle('active', wanted);
+
+        if (wanted && !structureProtocol) {
+            // Fetched when the layer is first asked for, not with the app: pmtiles and its
+            // inflate cost 8.7 kB gzipped, and most visits never press this. Held as a
+            // promise so two quick presses share one fetch, and the protocol is registered
+            // once per page -- maplibre keeps protocols globally and a second handler for
+            // the same scheme throws.
+            structureProtocol = import('pmtiles').then(({ Protocol }) => {
```
biodranik, line 220: A failed chunk load poisons the button for the rest of the page: the rejected promise stays in `structureProtocol`, `await structureProtocol` throws out of an async `onclick` (unhandled rejection, no message anywhere), and every later press hits the same cached rejection — button blue, layer never drawn, no retry. The repo's policy for a fetch that can fail is catch, log, degrade:

```ts
structureProtocol = import('pmtiles')
    .then(({ Protocol }) => { maplibregl.addProtocol('pmtiles', new Protocol().tile); })
    .catch(e => { structureProtocol = null; throw e; });
...
try { await structureProtocol; } catch (e) {
    console.error('Could not load the pmtiles protocol', e);
    button.classList.remove('active');
    shown = false;
    return;
}
```

---

src/map/map.ts
```diff
@@ -114,3 +126,174 @@ export async function loadSvgWithColors(svgSrc: string, colors: SelectorToColorP
...
+        // before it adds it -- so a press during startup used to leave the control holding
+        // a layer the map does not have, and the button dead from then on.
+        await loaded;
+        if (thisPress !== press) {
+            // Pressed again while this one was waiting; that press owns the outcome.
+            return;
+        }
+
+        wanted ? layerControls.addOverlayImmediate(spec) : layerControls.removeOverlayImmediate(spec);
+        if (wanted && map.getZoom() < STRUCTURE_ZOOM && map.getZoom() >= NEAR_ENOUGH_ZOOM) {
```
biodranik, line 253: Below `NEAR_ENOUGH_ZOOM` this produces exactly the dead button the new `.active` rule was added to avoid: the button turns blue, no tile exists at that zoom, nothing draws, and nothing says why. That's the app's own default view (z4 on desktop, z1 on mobile), so it's the first press a new user makes.

Cheapest fix that keeps the reasoning intact: leave the button disabled below `NEAR_ENOUGH_ZOOM` with a `title` saying "zoom in to see the stop structure", so the control is honest instead of silent.

---

src/uielements/selection-info.tsx
```diff
@@ -187,13 +207,75 @@ function MatchInfo({ datasetName, properties, geometry, idTags, reportRegion }:
...
+                    {minted.length > 0 && <>
+                        <div>{applied ? 'Split into' : 'Would be split into'} places the feed
+                            never named — these ids exist only in the published data, not in
+                            the feed:</div>
+                        <ul>{minted.map(part => <li key={part}><b>{part}</b></li>)}</ul>
+                    </>}
+                </>;
+            })()}
+        </div>;
+        })()}
```
biodranik, line 265: Two IIFEs nested inside JSX, the outer one de-indented back to column 8 to fit — the block reads as neither JSX nor a function. The file already has the shape that fixes it (`DatasetHelp`, `OsmElements`, `OsmListElement`): pull this out as

```tsx
function DissolutionPlan({ properties }: { properties: { [k: string]: any } }) { … }
```

with `applied`, `parts`, `minted`/`repointed` as plain `const`s at the top and a single `return`. `MatchInfo` then carries one line, and the same for the `notDissolved` block above. No behaviour change, ~40 lines of nesting gone.

(Minor, same block: `properties['dissolving']` vs `properties.gtfsStopId` — both styles appear within ten lines of each other.)

---

src/uielements/selection-info.tsx
```diff
@@ -187,13 +207,75 @@ function MatchInfo({ datasetName, properties, geometry, idTags, reportRegion }:
...
+                const partIds = parseJsonSafe<string[]>(parts, []) as string[];
+                // A minted id is the stop's own id plus the OSM feature it stands at; a
+                // re-pointed one is a stop the feed already has. The two look different
+                // because they are different: only the first is invented.
+                //
+                // The whole prefix, not the separator alone. Feed ids are free-form and do
+                // contain '#' -- this app encodes it in the URL hash for that reason -- so a
+                // contains-test would tell a mapper that their own stop `STN#3` exists only
+                // in the published data.
+                const mintedPrefix = `${properties['gtfsStopId']}#`;
```
biodranik, line 247: This block is rendered outside the `gtfsFeatures.length === 1` guard, so it also runs for a clustered record, where the ids live in `gtfsFeatures[]` and there is no top-level `gtfsStopId`. The prefix is then the literal `"undefined#"`, nothing matches it, and every part is printed under "the stops this feed already names as its parts" — the one caption the comment above says must not be wrong, since it tells a mapper an id exists in the feed when it was invented.

A guard costs one line and fails visibly instead:

```ts
if (!properties['gtfsStopId']) console.warn('dissolution parts without a stop id', partIds);
```

or skip the split and print one list when the id is missing.

---

src/app.css
```diff
@@ -293,3 +293,39 @@
...
+    border-left-color: #8b8b8b;
+    background: rgba(139, 139, 139, 0.08);
+}
+
+/* And for a stop it looked at and refused. Neither of the above: nothing is planned for it
+   and nothing is wrong with it, so it carries no warning colour at all. */
+.dissolving-plan.not-dissolved {
+    border-left-color: #9aa4ad;
+    background: rgba(154, 164, 173, 0.10);
+}
```
biodranik, line 322: `#8b8b8b` and `#9aa4ad` at 8 %/10 % on a 3px border are the same grey to the eye — two rules, two class names and two paragraphs of comment for one visual result. The distinction that actually reads is amber (acted) vs grey (not acted), which one modifier gives you. Suggest merging into `.dissolving-plan.not-acted` and dropping the second block; the words in the panel already say which of the two grey cases it is.

---

src/uielements/report-help-overlay.tsx
```diff
@@ -68,6 +68,20 @@ export function ReportHelpOverlay({ onClose }: { onClose: () => void }) {
...
+                    keeps whatever category matched it, so it is still shown and hidden by that
+                    category's checkbox. Where the tool measured such a stop without acting on
+                    it — a feed whose report says <b>analysis</b>, which is asked for per feed —
+                    it is marked <b>dissolvable</b> and drawn normally, because nothing was done
+                    to it. The selection panel says where its departures went, or would go, and
+                    names the rule where dissolution looked at a stop and left it whole. The
+                    report panel below says whether dissolution was switched on for the feed,
+                    only measured, or off — a feed with no marks and a feed nobody measured
+                    look the same otherwise.
+                </p>
```
biodranik, line 84: Proofreading, across all the user-visible strings this PR adds:

1. **"dealt to"** is card-game English and doesn't carry "reassigned" to a reader, least of all a non-native one. It's in the UI six times — "Visits dealt to", "Visits would be dealt to", "its departures are dealt to the places it stands for", "the stops it would deal to". **"moved to"** / **"reassigned to"** says it plainly.
2. **One noun, not three.** The help says *departures*, the panel says *visits*, the code comments say *calls* — all for the same thing. Pick one (*departures* reads best to a mapper) and use it everywhere.
3. **`dissolvable` / `analysis` in bold are pipeline vocabulary**, shown to a user who has never seen the report JSON. Either gloss them once or drop them — the sentence works as "the tool measured the stop but changed nothing".
4. **This paragraph is one ~110-word chain** with three subordinate asides ("— a feed whose report says analysis, which is asked for per feed —"). Every other paragraph in this overlay is two or three short sentences; matching that would roughly halve it.
5. `aChildsAnchorIsNotAmongItsFeatures` (selection-info.tsx:54) → "one of the stops it would deal to is not anchored to any of its features" is unreadable to the audience the map says it's for. Something like "one of the stops it would move departures to isn't attached to any OSM feature of this stop".
6. `map-tools.tsx`: **"PT structure"** — the rest of the panel spells things out (Help, Preview, Goto OSM); "Stop structure" fits and needs no abbreviation. And renaming **Map Style → Sat/Geo** is unrelated to this PR (and less clear for a control that cycles styles) — worth splitting out or dropping.

---

src/uielements/report.tsx
```diff
@@ -24,6 +24,14 @@ const GROUPS: { group: Group; title: string }[] = [
 
 const PREVIEW_COLOR = '#2c2ca5ff';
 
+// A stop whose calls this run dealt to the places it stands for is drawn grey rather than in
+// its match category's colour: it is still matched, still filtered by that category's
+// checkbox, and still in the data -- what moved is its departures. Only `dissolved` is drawn
+// this way. `wouldDissolve` and `dissolvable` describe a run that acted on nothing, so those
+// stops keep their category's own pin; the panel says what would have happened.
+const DISSOLVED_ICON = 'stop-dissolved';
+const DISSOLVED_COLOR = '#8b8b8b';
```
biodranik, line 33: Alternative worth weighing: grey *replaces* the category colour, so it spends the only channel the map has. A dissolved `clu`, a dissolved `mat` and a dissolved `hub` become indistinguishable, and `#8b8b8b` sits right next to `nos`'s black — the two states this map most needs to keep apart are the two that now look alike.

Keeping the category icon and adding dissolution as a second channel costs about the same:

```ts
'icon-opacity': ['case', ['boolean', ['get', 'dissolved'], false], 0.45, 1]
```

or a small circle layer beneath, filtered on `dissolved`. Pros: both facts readable at once, no extra icon image, and the `iconColors` special case at line 526 disappears. Cons: opacity is a weaker signal than colour on a busy map, a circle costs one more layer. Your call — but the current version is worth checking over a `clu`-heavy region before settling.

---

src/services/matchIndex.ts
```diff
@@ -162,6 +189,10 @@ export function parseIndex(tsv: string): ParsedIndex {
             continue;
         }
 
+        // Absent in a report from before the flags existed, which reads as a stop nothing
+        // was decided about -- true of every stop of a feed nobody analysed.
+        const flags = at['strategies'] !== undefined ? parseInt(c[at['strategies']], 10) : 0;
```
biodranik, line 194: `parseInt` returning `NaN` here is indistinguishable from `0`: `NaN & FLAG` is `0`, so a `strategies` column the pipeline ever writes non-numerically (a name list, hex, a quoted value) silently means "no stop in this region dissolves" — the exact silent-blank-map failure the header comment and the `skipped` counter were added to prevent. One line reuses the machinery already here:

```ts
const raw = at['strategies'] !== undefined ? c[at['strategies']] : '';
const flags = raw ? Number(raw) : 0;
if (!Number.isFinite(flags)) { skipped++; continue; }   // or count separately into indexError
```

Dropping a row may be too harsh for a flags column that isn't required — a separate counter folded into the `problems` list in `report.tsx` would say it without costing the stop.

---

biodranik (commented):
Reviewed the whole diff plus the callers it touches (`LayerControls`, `StopsLayer`, `parseIndex`, `MatchInfo`). CI green, no existing review threads. The dissolution model itself reads well — the "not a deleted stop" framing is right, and splitting minted vs re-pointed ids on the whole prefix rather than on `#` is the correct call for free-form feed ids.

Findings inline, most severe first:

1. **`dissolutionDeclined` is never read** — dead field + constant on every row of every region.
2. **`POI_RANK_FILTERS` hand-copies OpenFreeMap's filters** — `getStyle()` already returns them; ~12 lines and a drift risk removed.
3. **A failed `import('pmtiles')` kills the toggle permanently** — cached rejection, unhandled, no retry.
4. **CLAUDE.md documents `{region}/pt-structure.pmtiles`; the code fetches one global archive.**
5. **Below z8 the toggle turns blue and draws nothing** — the dead button the new `.active` rule exists to prevent, from the app's own default view.
6. **`"undefined#"` prefix** when a dissolution record has no top-level `gtfsStopId` (clustered records), mislabelling invented ids as feed ids.
7. Simplifications: extract the JSX IIFEs into a `DissolutionPlan` component; merge the two indistinguishable grey CSS variants.
8. Alternatives: dissolution as a second visual channel instead of overwriting the category colour; the POI filter applied in `resolveStyles` via the already-present-but-unused `mixinToStyle` rather than on every `style.load`.
9. Wording: "dealt to" throughout the UI, three nouns for one concept, and pipeline vocabulary shown to mappers — details on the help-overlay comment.

---

PR #164, opened 2026-09-13
Unmatched OSM

kiselev-dv:
Add an option to show GTFS stop at its matched location.

Add OSM Matches view, which shows matches from the OSM elements point of view. In other words Shows unmatched OSM stops in the vicinity of GTFS stop.


src/uielements/report.tsx
```diff
@@ -231,136 +315,119 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
...
+                geometry: { type: 'Point', coordinates: [lon, lat] },
+                properties: {
+                    name, lon, lat, flavour, seenBy, nearestM,
+                    osmFeatures: [{ id: osmId, lon, lat, tags: {} }],
+                },
+            }),
+            // No dataset: this is not one of the report's stop categories, and naming one
+            // would have the panel describe a match that does not exist.
+            reportRegion, idTags,
+        }, 'map-click');
```
biodranik, line 366: An OSM-only selection goes through `MatchInfo`, which assumes a GTFS stop behind it:

- `getGtfsFeatures` (selection-info.tsx:359) falls back to `[{id: properties.gtfsStopId, …}]`, so `gtfsFeatures.length === 1` and the panel prints `GTFS stop ID: **` (empty) and `GTFS stop code: N/A`.
- `AddOsmStopController id={properties.gtfsStopId}` is rendered with `id`/`code` both undefined. Pressing *Add OSM Stop* and clicking the map creates a node with `{[gtfsIdTag]: code || id}` → `ref:gtfs=undefined` (add-stop-controller.tsx:36-40), which then goes into the exported `.osm`.

Worth gating the GTFS blocks and the editor action in `MatchInfo` on an actual stop, e.g. `const hasGtfs = !!(properties.gtfsStopId || properties.gtfsFeatures)`.

Related: `flavour`, `seenBy` and `nearestM` are put into the properties here but nothing renders them — they are the only facts that make the circle worth clicking ("offered to 3 stops, nearest 12 m"). Either show them in the panel or drop them from the payload.

---

src/uielements/report.tsx
```diff
@@ -196,17 +213,84 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
...
+    const showingOsmMatches = osmMatchesOn && anchoredTotal > 0;
+
+    useEffect(() => {
+        setOsmMatchesAvailable(anchoredTotal > 0);
+        return () => setOsmMatchesAvailable(false);
+    }, [anchoredTotal, setOsmMatchesAvailable]);
+
+    // What the matcher looked at and did not use, drawn beside the stops it placed. Fetched
+    // only while osm matches is on: it is the largest file the report publishes, and a
+    // session that never looks at the osm-matches view should never pay for it.
```
biodranik, line 247: Outdated comment: the fetch is driven by `showUnmatchedOsm` — its own checkbox — not by the osm-matches switch. The dataset is deliberately independent of it (docs/component-tree.md says so), so this reads as a constraint that isn't there.

Same claim in two more places: the `UnassignedOsmLayer` doc ("while osm matches is on", report.tsx:693) and osmIndex.ts:4.

---

src/uielements/report.tsx
```diff
@@ -196,17 +213,84 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
...
+    const [unassignedRows, setUnassignedRows] = useState<OsmIndexRow[] | null>(null);
+    useEffect(() => {
+        if (!showUnmatchedOsm) return;
+        let cancelled = false;
+        loadUnmatchedOsmStops(reportRegion)
+            .then(rows => { if (!cancelled) setUnassignedRows(rows); })
+            .catch(e => {
+                console.error('Could not read osm-index.tsv.gz for', reportRegion, e);
+                setActionError(`Could not read osm-index.tsv.gz: ${e.message}`);
+            });
```
biodranik, line 258: Two small things:

1. The `.then` above checks `cancelled`, this doesn't — untick the checkbox while the fetch is in flight and a failure still raises a banner for a layer that is no longer on.
2. The file fetched is `osm-index-stops.tsv.gz`, not `osm-index.tsv.gz`; the error names the wrong one (also osmIndex.ts:80).

```suggestion
            .catch(e => {
                console.error('Could not read osm-index-stops.tsv.gz for', reportRegion, e);
                if (!cancelled) setActionError(`Could not read osm-index-stops.tsv.gz: ${e.message}`);
            });
```

---

src/services/osmIndex.ts
```diff
@@ -0,0 +1,106 @@
...
+
+        rows.push({
+            osmId: c[at['osm:id']],
+            lon,
+            lat,
+            flavour: at['flavour'] !== undefined ? c[at['flavour']] : '',
+            seenBy: at['gtfs_seen_by'] !== undefined ? parseInt(c[at['gtfs_seen_by']], 10) : 0,
+            nearestM: at['nearest_m'] !== undefined ? parseFloat(c[at['nearest_m']]) : NaN,
+            matched: 0,
+            anchored: parseInt(c[at['gtfs_anchored']], 10) || 0,
```
biodranik, line 101: `matched` is hard-coded to `0` although `gtfs_matched` is listed in `REQUIRED` — the column is demanded and never read, and the field's doc ("How many GTFS stops matched it, by any tier") is not what the code produces. `anchored` is parsed but no consumer reads it either (report.tsx only takes `osmId`, `lon`, `lat`, `flavour`, `name`, `seenBy`, `nearestM`).

This file is by definition the subset nothing matched, so the simplest fix is to drop both fields from `OsmIndexRow` and `gtfs_matched`/`gtfs_anchored` from `REQUIRED`. Also a stray double blank line at :87-88.

---

src/services/osmIndex.ts
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * The OSM stops and stations the matcher was offered and nothing matched.
+ *
+ * <p>Null where the region's report has no such file — one written before the matcher
+ * recorded its candidate pool. An absence, not a failure: the map then has nothing to draw
+ * and says so with an empty count.
+ */
+export async function loadUnmatchedOsmStops(region: string): Promise<OsmIndexRow[] | null> {
+    return cache[region] ??= fetchRegion(region);
```
biodranik, line 57: A failed fetch is cached as a rejected promise for the rest of the session, so unticking and re-ticking the checkbox never retries — the one recovery a user has. `routeVariants.ts` evicts on failure for exactly this reason; worth matching it:

```suggestion
    return cache[region] ??= fetchRegion(region).catch(e => {
        delete cache[region];
        throw e;
    });
```

---

src/services/osmIndex.ts
```diff
@@ -0,0 +1,106 @@
+// The report read from the OSM side: one row per feature the matcher was offered as a
+// candidate for some stop, and what the GTFS side made of it.
+//
+// Fetched only when osm matches is on, and once per region. It is the largest file the report
+// publishes — the pool is nearly twice the stop count — so it is written gzipped and never
+// loaded with the index: swiss-opendata is 11.6 MB of text and 3.3 MB on the wire.
```
biodranik, line 6: The header describes a file this module does not fetch. It fetches `osm-index-stops.tsv.gz` (the 67k-row subset, per the comment at :63-65), while "the largest file the report publishes … 11.6 MB of text and 3.3 MB on the wire" are the full pool's figures. And "only when osm matches is on" is the wrong trigger — it is the *Unmatched OSM stops* checkbox.

---

src/uielements/report.tsx
```diff
@@ -574,101 +683,115 @@ function StopsLayer({ layerKey, data, selectedCodes, onClick }: StopsLayerProps)
...
+            // @ts-ignore — the same immediate add the report's own overlays use, so the layer
+            // survives a base-style switch.
+            stylingControls.addOverlayImmediate(spec);
+            map.on('click', layerId, handleClick);
+            map.on('mouseenter', layerId, showPointer);
+            map.on('mouseleave', layerId, hidePointer);
+            // Under the stop pins: a stop the matcher did place must never be hidden by a
+            // feature it did not.
+            if (map.getLayer(layerId) && map.getLayer(`stops-${layerKey}`)) {
+                map.moveLayer(layerId, `stops-${layerKey}`);
```
biodranik, line 770: This ordering does not survive a base-style switch. `LayerControls.setBaseStyle` rebuilds the style by appending `this.overlays` in **registration** order, and the stops overlay registers on region load while this one registers when the checkbox is ticked — so after `cycleBaseStyle` the circles are appended after the stop symbols and the invariant the comment states ("a stop the matcher did place must never be hidden by a feature it did not") is gone.

Either re-apply the `moveLayer` on `styledata`, or give `OverlaySpecification` an optional `beforeId` that both `addOverlayImmediate` and `setBaseStyle` honour — the second is the only one that fixes it for every overlay.

---

src/uielements/report.tsx
```diff
@@ -231,136 +315,119 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
...
-            const lonlat = (selection.feature.geometry as { coordinates: number[] } & any)?.coordinates;
-            console.log('about to fly to', selection?.feature);
-            map.flyTo({ center: lonlat, zoom: 18, duration: 1 });
+            // Where the marker is, not where the feed put the stop: with osm matches on the
+            // two differ by up to 742 m on swiss-opendata, and the camera would land on empty
+            // map beside a stop whose panel is open.
+            const row = rowsById.get(selection.feature.properties?.gtfsStopId);
+            const drawnAt = showingOsmMatches && row?.anchorLon !== null && row?.anchorLat != null
+                ? [row!.anchorLon, row!.anchorLat]
+                : (selection.feature.geometry as { coordinates: number[] } & any)?.coordinates;
```
biodranik, line 420: `row?.anchorLon !== null` is `true` when there is no row at all (`undefined !== null`); only the loose `!= null` on the next test keeps this correct, and then `row!` asserts what the guard didn't establish. One spelling for both, and the `!`s go away:

```suggestion
            const row = rowsById.get(selection.feature.properties?.gtfsStopId);
            const drawnAt = showingOsmMatches && row?.anchorLon != null && row.anchorLat != null
                ? [row.anchorLon, row.anchorLat]
                : (selection.feature.geometry as { coordinates: number[] } & any)?.coordinates;
```

---

src/uielements/selection-info.tsx
```diff
@@ -300,11 +341,19 @@ function MatchInfo({ datasetName, properties, geometry, idTags, reportRegion }:
...
+ * fire. Anything derived from the elements takes the revision as its dependency.
+ */
 function useOsmFeatures() {
-    return useSyncExternalStore(
+    const revision = useSyncExternalStore(
         (sub) => OSM_DATA.subscribe(sub),
-        () => OSM_DATA.elements
+        () => OSM_DATA.revision
     );
+    return { elements: OSM_DATA.elements, revision };
```
biodranik, line 356: Simplification: the hook leaks `revision` to every caller, who must then remember to put it in each dep list — and the one caller already gets it half wrong, listing both `allOsmFeatures` (identity never changes, so it contributes nothing) and `osmRevision` at :473. Returning a snapshot whose identity *does* change makes the deps correct by construction:

```suggestion
function useOsmFeatures() {
    const revision = useSyncExternalStore(
        (sub) => OSM_DATA.subscribe(sub),
        () => OSM_DATA.revision
    );
    return useMemo(() => OSM_DATA.elements.slice(), [revision]);
}
```

Callers go back to `const allOsmFeatures = useOsmFeatures()` with plain `[allOsmFeatures, …]` deps (:421 and :473). A shallow copy of pointers per Overpass response is nothing beside the round-trip that produced it, and no future memo can forget the revision.

---

src/app.tsx
```diff
@@ -97,11 +96,49 @@ export const SelectionContext = createContext<SelectionContextT>({
...
   const [selectionSource, updateSelectionSource] = useState<SelectionSourceT>('app-init');
+  // Shared with the selection panel, which is the report's sibling rather than its child.
+  const [osmMatchesOn, setOsmMatchesOn] = useState(false);
+  const [osmMatchesAvailable, setOsmMatchesAvailable] = useState(false);
+
+  // A region with no anchors, or the report list, leaves nothing to show on osm matches -- and
+  // the switch outlives both, so it is cleared here rather than by whichever component noticed.
+  useEffect(() => {
+    if (!osmMatchesAvailable && osmMatchesOn) setOsmMatchesOn(false);
+  }, [osmMatchesAvailable, osmMatchesOn]);
```
biodranik, line 135: Three mechanisms guard the same thing: this effect, `showingOsmMatches = osmMatchesOn && anchoredTotal > 0` (report.tsx:238), and `OsmMatchesSwitch` returning `null` when unavailable. The last two already make the flag unreadable and the control unreachable, so this one only costs an extra render pass on every region switch — and dropping it is arguably nicer behaviour: moving between two regions that both have anchors would keep the switch where you left it instead of resetting it.

---

src/uielements/switch.tsx
```diff
@@ -0,0 +1,56 @@
+import { useContext } from "preact/hooks";
+
+import "./switch.css";
+import { OsmMatchesOptionsContext } from "../app";
```
biodranik, line 4: This makes a cycle: `app.tsx` → `selection-info.tsx` → `switch.tsx` → `app.tsx`. It works only because the context is read inside the component body rather than at module scope. Keeping `switch.tsx` to the generic `Switch` and putting `OsmMatchesSwitch` next to the context (in `app.tsx`, or its own file) breaks it and leaves a reusable control behind.

---

src/uielements/report.tsx
```diff
@@ -401,14 +466,36 @@ export function MatchReport({ reportRegion, reportData }: MatchReportProps) {
...
-                        ...selectedDatasets, preview: (e.target as HTMLInputElement).checked,
-                    })} />
-                <span className={'match-dataset'} title={'Show the stop positions and routes the matcher wrote'}>Preview</span>
+                <input className={'match-dataset-select'} type={'checkbox'} checked={showUnmatchedOsm}
+                    onChange={e => setShowUnmatchedOsm((e.target as HTMLInputElement).checked)} />
+                <span className={'match-group-title'}
+                    title={'OSM stops and stations the matcher was offered for some GTFS stop'
+                        + ' and nothing matched, and that nothing else in the pipeline claims.'
+                        + ' Shown beside any of the categories above; it moves nothing.'}>
+                    Unmatched OSM stops
```
biodranik, line 494: Proofreading: "the matcher was offered for some GTFS stop and nothing matched" reads as a garden path — *offered* wants *as a candidate for*, and the two clauses switch subject mid-sentence. "that nothing else in the pipeline claims" is also something the reader cannot check from the map.

```suggestion
                <span className={'match-group-title'}
                    title={'OSM stops and stations the matcher considered as candidates for'
                        + ' some GTFS stop and matched to none. Drawn beside any of the'
                        + ' categories above; it moves nothing.'}>
```

---

CLAUDE.md
```diff
@@ -51,15 +51,15 @@ Component/layout tree, side-panel tabs, and context consumers are documented in
...
-- Render-less overlay components follow one idiom: build the overlay spec, `mapLoaded.then(...)` guarded by a `subscription = { canceled, promiseFulfiled }` object, clean up in the effect teardown. See `StopsLayer`, `DatasetMapLayer` (preview), `RegionMarkersLayer`, `MatchArrowLayer`, `RoutesMap` (which instead keeps a persistent `routes` source and `setData`s into it), `HtmlMapMarker` (`editor/map-marker.tsx`).
+- `StopsLayer` (`report.tsx`) is **one** source/symbol layer for all of a region's stops; category toggles use `map.setFilter` on `subcategory`, and the osm-matches view swaps the geometry with `setData` from one of two collections built per region and kept — the source is created once per region, never per toggle. Icons are data-driven `stop-{code}` images recolored from `/stop-var.svg` by `loadSvgWithColors`, with `icon-opacity` fading the stops the anchoring refused. It also mutates the stored overlay spec's `filter` and `data` in place so base-style switches re-apply what is on screen.
```
biodranik, line 61: "with `icon-opacity` fading the stops the anchoring refused" — there is no `icon-opacity` anywhere in `src/`; `StopsLayer`'s layout has only `icon-image`, `icon-size` and `icon-allow-overlap`, and an unanchored stop is drawn at full opacity at its feed position. Either the fade was dropped or this sentence should go.

(The fade would be worth having, incidentally: with the switch on, the unanchored stops are the ones that did *not* move, and nothing on the map distinguishes them from stops anchored exactly where their feed puts them — only the panel says which, via `notAnchored`.)

---

biodranik (commented):
Re-review of `eec901b` (merged — these are follow-ups). Reviewed as architect, frontend, performance, OSM and GTFS: the two-collection `setData` design, the `revision` counter and the on-demand gzipped index are all sound; findings are inline.

The one that matters: clicking an unmatched OSM circle opens the GTFS stop panel, which offers **Add OSM Stop** and mints a node tagged `ref:gtfs=undefined`. The rest are comment/doc drift, dead fields in `OsmIndexRow`, a cached rejected promise, and layer order lost on a base-style switch.

**Alternative worth weighing** — the unmatched-OSM dataset is a 3.3 MB gzipped TSV parsed into a 67k-feature GeoJSON source held whole in memory (germany-local). The pipeline already publishes `pt-structure.pmtiles`; putting this layer in a PMTiles archive would make it viewport- and zoom-lazy, with no `DecompressionStream`, no full-region parse and no per-region memory cost. Cons: pipeline work, a second format for one layer, and the current path is already lazy behind a checkbox. Probably not worth it for one country — worth it if this grows to the whole planet or gains per-feature styling.

Two things I checked and found fine: `manualChunks: { pmtiles: [...] }` does only name the chunk that `await import('pmtiles')` in map.ts already splits out (the comment is accurate), and `MatchReport` is keyed by region in `report-selector.tsx`, so `unassignedRows` cannot survive a region switch.

---

PR #165, opened 2026-09-14
Fix and add tags for OSM changesets

kiselev-dv:
* Fix nodes for edited ways
* Add changeset tags


src/uielements/editor/changes.tsx
```diff
@@ -59,24 +60,41 @@ export function Changes({ osmData }: ChangesProps) {
...
             .map(([k, v]) => ({ tag: { _attr: { k, v } } }));
 
-        const { type, tags, ...attr } = element;
+        const { type, tags, ...rest } = element;
+        const { nodes, ...attr } = rest as Record<string, any>;
 
         // @ts-ignore
         attr['action'] = 'modify';
         // @ts-ignore
         attr['version'] = '1';
```
biodranik: `version` is hard-coded to `1`, throwing away the real version that Overpass (`out meta`) and `api.openstreetmap.org/…/nodes.json` return in the element, while `timestamp`/`changeset` of the true version are kept in the same attribute list. JOSM uploads the version from the file, so every element currently at version > 1 — i.e. most edited stops — gets a 409 version mismatch.

Rendering the current code for a way at version 7 gives:

```xml
<way id="1420729449" version="1" timestamp="2024-01-02T03:04:05Z" changeset="1234" … action="modify">
```

Emit the element's own version instead, `0` for the locally created negative-id nodes (JOSM's `AbstractReader.parseVersion` requires version > 0 only for `id > 0`, and JOSM itself writes `version='0'` for new objects):

```ts
version: element.version ?? 0,
```

That needs `version?: number` on `OSMNode`/`OSMWay`/`OSMRelation` in `services/OSMData.types.ts` (verified: `yarn build` passes with it).

---

src/uielements/editor/changes.tsx
```diff
@@ -59,24 +60,41 @@ export function Changes({ osmData }: ChangesProps) {
...
 
+        const children: any[] = [...tagElements];
+
+        if (element.type === 'way' && Array.isArray(nodes)) {
+            const ndElements = nodes.map((nodeId: number) => ({ nd: { _attr: { ref: nodeId } } }));
+            children.unshift(...ndElements);
+        }
```
biodranik: Building the attributes explicitly is shorter and drops the `as Record<string, any>` cast, the `any[]`, and the `Array.isArray(nodes)` guard — the `OSMWay` branch of the union already guarantees `nodes`, and the repo's policy is to fail visibly rather than guard our own data. It also fixes the version above and keeps Overpass meta (`user`, `uid`, `timestamp`, `changeset`) out of the file, which JOSM does not need:

```suggestion
        // Only what OSM XML defines for an element; a wrong version makes the
        // upload conflict, and new elements (negative id) have no version yet.
        const attr: Record<string, string | number> = {
            id: element.id,
            version: element.version ?? 0,
            action: 'modify',
        };

        if (element.type === 'node') {
            attr.lat = element.lat;
            attr.lon = element.lon;
        }

        const children = element.type === 'way'
            ? [...element.nodes.map(ref => ({ nd: { _attr: { ref } } })), ...tagElements]
            : tagElements;
```

Also note the two `// @ts-ignore`s are dead now: after `rest as Record<string, any>`, `attr` has an index signature, and `tsc` reports both as unused when switched to `@ts-expect-error`.

(Validated: with `version?: number` added to the element types, `yarn build` passes and the rendered XML is unchanged apart from version/meta.)

---

src/uielements/editor/changes.tsx
```diff
@@ -59,24 +60,41 @@ export function Changes({ osmData }: ChangesProps) {
...
+    const changesetTags = [
+        { tag: { _attr: { k: 'created_by', v: 'gtfs-osm-matcher.organicmaps.app' } } },
+        { tag: { _attr: { k: 'hashtags', v: '#gtfs;#gtfs-osm-matcher.organicmaps.app' } } },
+    ];
+
+    if (region) {
+        changesetTags.push({ tag: { _attr: { k: 'data_used', v: region } } });
+    }
+
+    return xml({ osm: [{ _attr: { version: "0.6", generator: "osm-gtfs" } }, ...changesetTags, ...xmlNodes] }, { declaration: true });
```
biodranik, line 96: These tags land as `<tag>` children of `<osm>`, which JOSM silently ignores, so the changeset half of this PR has no effect on upload:

- `OsmReader.parseOsm()` switches on `bounds | node | way | relation | changeset | remark` and routes everything else to `parseUnknown()` — logging *"Undefined element ''tag'' found in input stream. Skipping."*
- `<changeset>` is no workaround either: `parseChangeset(uploadChangesetId)` only reads its tags when the element's `id` equals the changeset being uploaded.

Rendered output today:

```xml
<osm version="0.6" generator="osm-gtfs"><tag k="created_by" …/><tag k="hashtags" …/>…
```

Alternatives that actually work:
1. Show the intended comment/hashtags as copyable text in the Changes tab, next to the download button — the user pastes them into JOSM's upload dialog. Smallest change, no file-format assumptions.
2. JOSM remote control does support prefilling them, but only for a URL JOSM can fetch: `/import?url=…&changeset_tags=comment=…|hashtags=…` (`ImportHandler` → `LoadAndZoomHandler.parseChangesetTags`). A `blob:` download can't be used that way, so this needs the file hosted somewhere first.

If the tags stay for other consumers, three things about the values:
- `created_by` is set by the uploading editor itself (JOSM overwrites it), so it never reaches the changeset from here.
- `#gtfs-osm-matcher.organicmaps.app` is not a usable hashtag — changeset hashtag tooling parses `#[\w-]+`, so it truncates at the first dot. `#gtfs;#organicmaps` reads better and is what people can search for.
- `data_used` gets the internal region slug of whichever region is open at download time, while `OSM_DATA.changes` is a global that is never cleared on region switch — one file can carry edits made across several regions, and would then be labelled with just the last one.

---

biodranik (commented):
The `<nd>` fix is the right one — without it a modified way uploads with an empty node list.

Three points inline: the changeset tags are dropped by JOSM's parser (evidence and alternatives in the comment), the hard-coded `version="1"` conflicts on upload, and the attribute building can be a good deal shorter and fully typed. All checked against the code and against JOSM's `OsmReader`/`AbstractReader`; the suggested rewrite builds (`yarn build`).

---

src/uielements/editor/changes.tsx
kiselev-dv: For now I will leave this as is. I don't have much of a control over JOSM and what it is ignoring. But this would be useful in the future for uploading changesets directly to OSM without JOSM.

PR #166, opened 2026-09-14
Update maplibre-gl to v6

kiselev-dv:
* Update dependency reported by dependabot
* Resolve build issues caused by update

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "serve-data": "http-server ../gtfs-parser/data/match -p 8801 --cors"
   },
   "dependencies": {
-    "maplibre-gl": "^5.10.0",
+    "maplibre-gl": "^6.4.1",
     "pmtiles": "^4.5.0",
     "preact": "^10.27.3"
   },
@@ -18,7 +18,7 @@
     "@babel/core": "^8.0.1",
     "@preact/preset-vite": "^2.10.2",
     "@types/node": "^25.2.0",
-    "baseline-browser-mapping": "^2.9.15",
+    "baseline-browser-mapping": "^2.11.0",
     "typescript": "~5.9.3",
     "vite": "^7.3.5",
     "wrangler": "^4.61.1"
--- a/yarn.lock
+++ b/yarn.lock
@@ -1113,22 +1113,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@mapbox/geojson-rewind@npm:^0.5.2":
-  version: 0.5.2
-  resolution: "@mapbox/geojson-rewind@npm:0.5.2"
-  dependencies:
-    get-stream: "npm:^6.0.1"
-    minimist: "npm:^1.2.6"
-  bin:
-    geojson-rewind: geojson-rewind
-  checksum: 10c0/631f89ba5b656cb1e02197c242b231f98da0afb96815fa26481497176d6bd5f2aac77af4950da91c954094694acbc26382bd3d38146705737e8ff06442d95a12
-  languageName: node
-  linkType: hard
-
-"@mapbox/jsonlint-lines-primitives@npm:^2.0.2, @mapbox/jsonlint-lines-primitives@npm:~2.0.2":
-  version: 2.0.2
-  resolution: "@mapbox/jsonlint-lines-primitives@npm:2.0.2"
-  checksum: 10c0/5814e42fc453700132f93ea742aabcef9a3c98d9bf17d4c1106f82d1dcd91bbc93052e66e29014323b9b2a41b020c743d897e4a96cc4ed2f734482d587d8c2b2
+"@mapbox/jsonlint-lines-primitives@npm:^2.0.3":
+  version: 2.0.3
+  resolution: "@mapbox/jsonlint-lines-primitives@npm:2.0.3"
+  checksum: 10c0/e114b27757a21ccbf20e3c99761e8b48d2527cb321f47e5a4d9aa4b5781b16f15b1f6f4ab678a4329f93668956281a05b56387331dc0260981a2ef695414fc1d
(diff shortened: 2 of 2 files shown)


src/uielements/editor/map-marker.tsx
```diff
@@ -46,8 +46,8 @@ export function HtmlMapMarker({ name, lat, lon, className, mouseEvents }: HtmlMa
         const { onClick, onHoverUpdate } = mouseEvents || {};
 
         m.on('click', () => onClick?.());
-        m.on('mouseenter', () => onHoverUpdate?.(true));
-        m.on('mouseleave', () => onHoverUpdate?.(false));
+        m.getElement().addEventListener('mouseenter', () => onHoverUpdate?.(true));
+        m.getElement().addEventListener('mouseleave', () => onHoverUpdate?.(false));
```
biodranik, line 50: These three listeners are dead code — no caller ever passes `mouseEvents`. All three `HtmlMapMarker` call sites (`selection-info.tsx:186`, `:413`, `:424`) omit the prop, so `onClick`/`onHoverUpdate` are always `undefined`. Marker highlighting is driven the other way round: `OsmListElement` hover sets `highlightId`, which flows back as the `highlight` class.

So instead of porting the handlers to DOM listeners, the whole prop can go — the `MouseEventsHandlers` import, the prop, the destructure, the three listeners and the `mouseEvents` dep:

```suggestion
```

together with:

```diff
-import type { MouseEventsHandlers } from "../selection-info";
 import { useContext, useEffect } from "preact/hooks";
@@
     className?: string;
-
-    mouseEvents?: MouseEventsHandlers;
 }
-export function HtmlMapMarker({ name, lat, lon, className, mouseEvents }: HtmlMapMarkerProps) {
+export function HtmlMapMarker({ name, lat, lon, className }: HtmlMapMarkerProps) {
@@
-    }, [map, name, lat, lon, className, mouseEvents]);
+    }, [map, name, lat, lon, className]);
```

`MouseEventsHandlers` stays exported for `OsmListElement`, which does use it. Verified: `tsc -b` and `vite build` pass with this removal.

Worth noting for the record that `m.on('mouseenter'|'mouseleave')` never fired in v5 either — `Marker` only ever emitted `dragstart`/`drag`/`dragend`/`click` (`MarkerEventType` in v6's d.ts). v6 just made it a type error rather than a silent no-op.

---

src/map/map.ts
```diff
@@ -1,4 +1,5 @@
-import maplibregl, { Map } from "maplibre-gl";
+import * as maplibregl from "maplibre-gl";
+import type { Map } from "maplibre-gl";
```
biodranik, line 2: v6 really did drop the default export, so this had to change — but only two symbols are used from the namespace (`maplibregl.Map` on line 18 and `maplibregl.addProtocol` on line 222), so one named import does it, and matches how the other seven maplibre imports in `src/` are written:

```suggestion
import { Map, addProtocol } from "maplibre-gl";
```

with `new Map({…})` and `addProtocol('pmtiles', …)` at the two call sites. `Map` covers both the value and the five type positions, so the separate `import type` line goes away. Verified: `tsc -b` and `vite build` pass; bundle size unchanged (the maplibre chunk is manual, so the namespace import costs nothing — this is about consistency, not size).

---

src/uielements/selection-info.tsx
```diff
@@ -2,6 +2,7 @@ import "./selection-info.css";
 
 import { useCallback, useContext, useEffect, useMemo, useState } from "preact/hooks";
 import { SelectionContext, type SelectionT } from "../app";
+import type { Geometry } from "geojson";
```
biodranik, line 5: `geojson` is not a declared dependency — this resolves only because `@types/geojson` is a dependency of `maplibre-gl` that the node-modules linker happens to hoist to the top level (confirmed: it is in `yarn.lock` only as a transitive of maplibre-gl/style-spec, and `node_modules/@types/geojson` exists without ever being asked for). A `yarn dedupe`, a nested install, or maplibre dropping that dep breaks `tsc -b` with no change on our side.

Since `geometry` is just `selection?.feature.geometry` passed down (line 68), taking the type from the value's actual source needs no import and no new dependency — `SelectionT` is already imported on line 4:

```suggestion
```

and on line 117:

```diff
-    geometry: Geometry | undefined
+    geometry: SelectionT['feature']['geometry'] | undefined
```

Verified: `tsc -b` and `vite build` pass. (The alternative — adding `@types/geojson` to `devDependencies` — also works, but a direct dep just to name a type the app already has a handle on cuts against the repo's "no new dependencies" rule.)

For context on why the old `GeoJSON.Geometry` stopped resolving at all: `tsconfig.app.json` sets `"types": ["vite/client"]`, which switches off automatic `@types/*` inclusion, so the UMD global only ever existed because v5's d.ts pulled `geojson` into the program. v6's no longer does.

---

biodranik (commented):
Reviewed post-merge (review was requested after the merge). The v6 migration itself is correct and complete: v6 exports no default and `Marker` only emits `dragstart`/`drag`/`dragend`/`click`, so both breakages were real, and nothing else in `src/` touches an API v6 changed (checked `queryRenderedFeatures`, `easeTo`/`flyTo`/`fitBounds`, `addImage`, `get/setStyle`, `setFilter`, `addProtocol` against the 6.4.1 d.ts). `yarn install --immutable && tsc -b && vite build` all pass locally.

Three cleanups in the inline comments, each verified to build. One thing not in the diff:

- `CLAUDE.md:13` still reads "Preact 10 + Vite 7 + TypeScript 5.9 (strict) + **MapLibre GL 5**" — stale as of this PR; worth a one-word follow-up so the agent guidance does not point at the old major.

---

PR #167, opened 2026-09-14
Bundle the MapLibre worker with Vite

Uagubkin:
MapLibre 6 resolves its worker next to the bundled library chunk by default. That file is not emitted by Vite, so production receives the SPA HTML fallback instead and vector maps never finish loading.

Import the worker through Vite's worker pipeline and configure its generated URL before creating the map.


src/map/map.ts
```diff
@@ -1,10 +1,13 @@
 import * as maplibregl from "maplibre-gl";
+import workerUrl from "maplibre-gl/dist/maplibre-gl-worker.mjs?worker&url";
 import type { Map } from "maplibre-gl";
 import 'maplibre-gl/dist/maplibre-gl.css';
 import { DATA_BASE_URL } from "../config";
 import { satMapStyle } from "./styling";
 import { LayerControls } from "./layers-controls";
 
+maplibregl.setWorkerUrl(workerUrl);
```
biodranik, line 9: Worth a comment — this is the kind of bare side-effect line a later reader deletes as redundant, and the two facts that make it necessary are both invisible from the code:

```suggestion
// MapLibre 6 resolves its worker with `new URL('./maplibre-gl-worker.mjs', import.meta.url)`,
// a form Rollup does not analyse, so Vite never emits the file. `?worker&url`, not `?url`:
// the shipped worker imports `maplibre-gl-shared.mjs` relatively, so a verbatim copy 404s in turn.
maplibregl.setWorkerUrl(workerUrl);
```

The second half is the part that bites if anyone "simplifies" the query string later — `maplibre-gl-worker.mjs` opens with a bare `import{...}from"./maplibre-gl-shared.mjs"`, so plain `?url` would just move the 404 one file along.

Module scope is also the right place (it has to run before the first `new Map()`, and `createMap` is the only thing that constructs one) — no change needed there.

---

biodranik (commented):
The diagnosis and the fix both hold up — I reproduced the bug and the repair in a real browser rather than by reading.

On `main`, the built page requests `/assets/maplibre-gl-worker.mjs` and gets **404**; with this branch it gets **200** and the dispatcher comes up with a live actor. The cause is exactly as described: `defaultWorkerUrl()` in 6.4.1 is `new URL('./maplibre-gl-worker.mjs', import.meta.url)`, which Rollup does not statically analyse, so the file is never emitted. `setWorkerUrl` is also the only override v6 still exposes — `workerClass` is gone — so there is no neater hook to reach for. `tsc -b` and `vite build` pass.

One thing to note but not act on: the emitted worker is 469 kB raw / 131 kB gzipped, because Vite inlines `maplibre-gl-shared` into it while the main chunk keeps its own copy. That duplication is unavoidable — a worker has its own module graph — and it only loads for people who render a map, i.e. everyone here.

**One follow-up, outside the diff:** `optimizeDeps.exclude: ['maplibre-gl']` in `vite.config.ts` (added in 43c0c49, "Fix maplibre-gl v6 build") becomes dead config once this lands, and I would drop it in the same PR.

It exists only as a dev-side workaround for this same root cause: with it removed on `main`, the dev server requests `/node_modules/.vite/deps/maplibre-gl-worker.mjs` → 404, because pre-bundling moves the module and `import.meta.url` follows it. Excluding maplibre kept it at its real path so the relative guess still landed. Once `setWorkerUrl` is set the default resolution is never consulted, so the exclusion has nothing left to protect.

Verified on this branch with the exclusion removed: dev serves the worker (200, one actor) with maplibre pre-bundled again — fewer dev requests and a faster cold start — and the production build is byte-identical, same asset hashes, since `optimizeDeps` is dev-only. Removing it also spares the next reader a piece of unexplained config whose reason will have evaporated.

---

biodranik (approved):
Thanks for the fix! Claude is nitpicking, need to tame it.

